PR #1, opened 2016-08-03
Negative margins are allowed

lesleh:
[The spec](https://www.w3.org/TR/CSS2/box.html#margin-properties) states that negative margins in CSS are valid.

> Negative margin values are allowed, but there may be implementation-specific limits.


aeschli:
Thanks @lesleh !

PR #3, opened 2016-10-20
Support !important in Variables (fix vscode#14120)

mrmlnc:
Source: https://github.com/Microsoft/vscode/issues/14120
## Before

image
## After

image


aeschli:
Thanks @mrmlnc !

PR #11, opened 2016-11-15
Fully support parsing Custom Property values.

### LINKED ISSUE #7: [feature request] Support CSS Custom Properties
[CSS Custom Properties](https://drafts.csswg.org/css-variables/) are a relatively new standard. Work is [in progress to support them in Edge](https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6261292-css-variables), and they're shipping today in [Chrome, Safari, and Firefox](http://caniuse.com/#feat=css-variables).

Custom property usage is fairly straightforward, just an expression like `var(--prop-name)` or `var(--prop-name, exprAsFallbackValue)`. The main tricky bit is that the [allowed syntax for custom property values](https://drafts.csswg.org/css-variables/#syntax) is extremely permissive, allowing stuff like:

```css
body {
  --js-might-read-this: if(x > 5) this.width = 10;

  --my-theme: {
    color: red;
    background: white;
  };
}
```

The latter style does see real world usage today to implement mixins, and is reported as a syntax error in vscode as of v1.7.1.

rictic:
Fixes #7

This adds a new method for parsing css custom property values. It's written somewhat unusually due to the exceptionally broad allowed syntax for custom property values.

### DIFF
--- a/src/parser/cssNodes.ts
+++ b/src/parser/cssNodes.ts
@@ -385,6 +385,7 @@ export class Nodelist extends Node {
 export class Identifier extends Node {
 
 	public referenceTypes: ReferenceType[];
+	public isCustomProperty = false;
 
 	constructor(offset: number, length: number) {
 		super(offset, length);
@@ -590,6 +591,10 @@ export class Property extends Node {
 
 	public getName(): string {
 		return this.getText();
+	}
+
+	public isCustomProperty(): boolean {
+		return this.identifier.isCustomProperty;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


abdonrd:
Thanks @rictic!

aeschli:
@rictic I merged your pull request manually. Thanks a lot!

Your implementation makes sense as it is what the spec describes. In the future we have to see how we can give better support to the definitions in the custom property set. I hope the spec gets more specific...

Issue #16, opened 2016-11-27
Completion for Mixins after @include


mrmlnc:
Failed test:

```ts
testCompletionFor('@mixin mixin($a: 1, $b) { content: $a + $b; } @include m|', {
	items: [
		{ label: 'mixin' }
	]
}),
```

I'll create PR for this case 😈

mrmlnc:
Closed by #17

PR #17, opened 2016-11-27
Add Mixin proposals

mrmlnc:
Source: #16

This PR adds:

  * Proposals for Mixins only after `@include` keyword

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -9,7 +9,7 @@ import {Symbols, Symbol} from '../parser/cssSymbolScope';
 import * as languageFacts from './languageFacts';
 import * as strings from '../utils/strings';
 import {findFirst} from '../utils/arrays';
-import {TextDocument, Position, CompletionList, CompletionItemKind, Range, SnippetString} from 'vscode-languageserver-types';
+import {TextDocument, Position, CompletionList, CompletionItem, CompletionItemKind, Range, SnippetString} from 'vscode-languageserver-types';
 
 export class CSSCompletion {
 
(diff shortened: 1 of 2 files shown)


mrmlnc:
Hmmm, I don't see ability to handling nested selectors for Mixins in Less.

```less
.mixin(@a: 1, @b) { content: @a + @b; }; .| // SimpleSelector
.mixin(@a: 1, @b) { content: @a + @b; }; .test { .|} // Declarations
.mixin(@a: 1, @b) { content: @a + @b; }; .test { .| // Nothing
```

aeschli:
@mrmlnc Thank a lot!
I had recreate the commit to merge the changes. Also I made a few (mostly cosmetic) changes.

mrmlnc:
@aeschli,

Yeap, this is good 👍. But what to do with Mixins in Less? We have three contexts and one of them does not exist in the AST.

aeschli:
For LESS we need to override 'parseRuleset' to detect rulesets with selectors that are valid mixin declarations and annotate the selector identifier.
On the reference side, we get MixinReference Nodes already for .abc and #abc. What is the 3rd context?

mrmlnc:
Now AST contains all Mixins in the Less file. Sorry, I was not entirely correct.

The problem is that Mixin can be called in different contexts:

  1. `SimpleSelector`

  ```less
  .mixin() {}
  .m|
  ```

  2. `Declarations`

  ```less
  .mixin() {}
  .test {
    .m|
  }
  ```

  3. Just case, not context.

  ```less
  .mixin() {}
  .test {
    .m|
  // without }
  ```

PR #18, opened 2016-11-29
Fix compilation error due to type mismatch

zeusk:
Changes to be committed:
	modified:   src/services/cssNavigation.ts

The switch to vscode-languageserver-types v3.0.1-alpha.2 results in the following error:

```
src/services/cssNavigation.ts(147,10): error TS2322: Type '{ changes: { [x: string]: TextEdit[]; }; }' is not assignable to type 'WorkspaceEdit'.
  Types of property 'changes' are incompatible.
    Type '{ [x: string]: TextEdit[]; }' is not assignable to type 'TextDocumentEdit[]'.
      Property 'length' is missing in type '{ [x: string]: TextEdit[]; }'.
```

This commit fixes the above.

### DIFF
--- a/src/services/cssNavigation.ts
+++ b/src/services/cssNavigation.ts
@@ -6,7 +6,7 @@
 
 import * as nodes from '../parser/cssNodes';
 import {TextDocument, Range, Position, Location, DocumentHighlightKind, DocumentHighlight,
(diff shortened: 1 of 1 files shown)


aeschli:
Thanks @zeusk. I fixed this by requiring exact versin v3.0.1-alpha.2 for now (the problem is that there's a change in the latest v3.0.1-alpha). I'll migrate to the latest at some point (will be a ripple in other dependencies as well)

PR #20, opened 2017-02-04
Support unprefixed hyphens

Tyriar:
Fixes Microsoft/vscode#19866 (after dep update)

---

Tested by making the same change to the `extensions/css/server/node_modules/vscode-css-languageservice/lib/data/browsers.js` in a VS Code workspace.

image

Support data sourced from https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#Browser_compatibility


aeschli:
`browser.js` is generated from https://github.com/Microsoft/vscode-css-languageservice/blob/master/build/css-schema.xml.
So the new attribute should be added there, and `npm run update-data` will then update browser.js

Tyriar:
@aeschli updated the browsers supported and regenerated, I don't need to change the version attr or anything do I?

aeschli:
@Tyriar Awesome, thanks a lot. I will bump the version and update the dependency in vscode in the coming days.

PR #23, opened 2017-02-16
Change sort order using splittedByHyphenCount

katainaka0503:
Related to Microsoft/vscode#20738.


aeschli:
Sorry I don't want push this PR. To me your suggested change is too specific to your problem and will cause behaviour that is unpredictable users.

PR #28, opened 2017-05-16
Update css-schema.xml to support CSS Grid

lannonbr:
Fixed Microsoft/vscode#13595.


aeschli:
Thanks @lannonbr !
As mentioned there is the danger that these changes get overwritten once we connect to the latest data sources. This is not planned for comming next months tough.

karocksjoelee:
I am using the latest version of vs code , does this updated ?? 
cause I am still getting [css] unknown property warning

PR #29, opened 2017-05-16
add: flow-root for display property completion

praveenpuglia:
`flow-root` is now a supported value for display which essentially replaces the clearfix hack. It's available in Chrome 58 and FF53 onwards.

It didn't appear in VSCode completion, so I thought I would make a PR here.

### DIFF
--- a/src/data/browsers.js
+++ b/src/data/browsers.js
@@ -2221,6 +2221,10 @@
 						"browsers": "O12.1"
 					},
 					{
+						"name": "flow-root",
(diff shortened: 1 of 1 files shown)


aeschli:
@praveenpuglia Hi Praveen, `browsers.js` is generated from `css-schema.xml`. I made your suggested change there, regenerated and pushed. I hope that's ok for you. Thanks for the PR!

praveenpuglia:
I didn't know that. Thanks for updating in right place. :)

PR #30, opened 2017-05-16
text-decoration-color update

ubershmekel:
Safari 7.1 supports it with `-webkit`.

https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color


aeschli:
Thanks @ubershmekel !

PR #35, opened 2017-07-07
grid-row-start property browser support info

praveenpuglia:
CC : https://github.com/Microsoft/vscode/issues/27590 

- One thing I noticed is that even though css-schema.xml has the description for the span value, in built browsers.js it's not showing up. @aeschli - please help with this.


praveenpuglia:
@aeschli - can we also publish a newer version to npm after this is merged? That way vscode will get the new support info automatically in next build.

aeschli:
Thanks @praveenpuglia !

Yes, I'll create a version and update VSCode so this gets in the next release of VSCode.

PR #37, opened 2017-07-16
add system font stack in font-family autocomplete list

praveenpuglia:
- add: npm-debug.log to .gitignore

P.S - Do I need to create an issue for this?

## Why System Font Stack?
When working on electron projects and web apps, it's often desired for them to have the system native font so they can blend in the environment they are being operated.

Links
- https://css-tricks.com/snippets/css/system-font-stack/
- https://medium.com/@praveenpuglia/tiny-css-for-nativity-f9a4b7f0148c

## Could we use a different variant of the stack? 
May be. May be not. We can discuss this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 lib/
 node_modules/
 coverage/
+npm-debug.log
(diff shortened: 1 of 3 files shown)


aeschli:
Thanks @praveenpuglia

Issue #32, opened 2017-06-07
Provide support for non standard CSS properties and restrictions in VSCode


SinZ163:
Hi, I've been using VSCode a bunch for Dota 2 Custom Games, and it does the job great, except when it comes to how Valve have altered CSS to suit their needs.

A list of their properties can be found [here](https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/Panorama/CSS_Properties), and having the ability to enhance / replace the current CSS Schema with a new properties and restrictions, such as `hue-rotation` and hex based color inputs accepting 8 hexadecimal characters for color (`#333333b0`), as opposed to the normal 6 (`#FF0000`). 

Note I'm not asking to have Valves CSS properties being a part of VSCode, rather providing the infrastructure to allow users to enhance the existing language service with new properties or features.

praveenpuglia:
I think this can be a plugin of some sort.

aeschli:
Yes, the idea of Microsoft/vscode-css-languageservice is that it is to be resused for css like languages. The 8 digit hex colors might come soon for css as well, so I'm happy to take a PR that adds that.
I'm also open to make the properties configurable

praveenpuglia:
I started working on a branch but I am not sure how do I test my changes in the editor. 

What I can't figure out is...
- How to get tell the editor that it's a valid color and don't show red underline.
- How to get the editor to show a box that shows the color beside the color definition.

https://github.com/praveenpuglia/vscode-css-languageservice/commit/0addc3d91f46eabaca19de22e3f63e08ec2ab622

aeschli:
The fix in parser. _parseHexColor should fix the validation as well as making the color to reach the client. But on the css client side you'll have to turn the color into a valid css color that Chromium understands: https://github.com/Microsoft/vscode/blob/master/extensions/css/client/src/colorDecorators.ts#L114 

To run VSCode with your version of  checked out in the same folder
- have the vscode repo and the vscode-css-languageservice checked out in the same folder
- see https://github.com/Microsoft/vscode/wiki/How-to-Contribute for instructions how to selfhost vscode
- in the vscode workspace, cd to `extensions/css/server/` and run `npm run install-service-local` to install your local vscode-css-languageservice node modules

praveenpuglia:
Alright! So I think I got a color box to appear beside a color but I am still failing validation.

I still get the validation error "Hex colors must have three or six hex numbers". Guess I have to change the validation here?

[...]

PR #38, opened 2017-07-17
CSS Colors Level 4 - 4 & 8 digit hex values

praveenpuglia:
fix: #32 

Note.
This doesn't show any color in small color box in the editor beside the color definition because css itself doesn't support it at the moment.

We could get around that problem by detecting and converting `#rgba` or `#rrggbbaa` values to `rgba()` values before they are added to `css/client/src/colorDecorators.ts`.

@aeschli - please correct me if I missed something. Very new to the source code.

### DIFF
--- a/src/parser/cssParser.ts
+++ b/src/parser/cssParser.ts
@@ -1217,7 +1217,7 @@ export class Parser {
 	}
 
 	public _parseHexColor(): nodes.Node {
-		if (this.peekRegExp(TokenType.Hash, /^#[0-9A-Fa-f]{3}([0-9A-Fa-f]{3})?$/g)) {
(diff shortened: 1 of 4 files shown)


aeschli:
Looks good. Can you also provide a fix for the color decorations (in vscode)? As you write, it needs to be detected and converted to 'rgba(...)' There is should also be fixed at https://github.com/Microsoft/vscode/blob/master/extensions/html/client/src/colorDecorators.ts#L114

praveenpuglia:
@aeschli - I am working on a fix. Thinking of adding a converter function in the `colorDecorators.ts` itself. Is there a better suited for these sort of helpers?

praveenpuglia:
@aeschli - made this commit to my own fork. Can you please check it once?
https://github.com/praveenpuglia/vscode/commit/434beabd8e4799393a027d1134897135379f7705

Also, couldn't understand where do I write tests for it.

egamma:
@praveenpuglia fyi, we have vacation time over hear so response time will we slower.

praveenpuglia:
@egamma - sure. Have a great one! 👍 😊

PR #41, opened 2017-08-11
browser support updates to some properties

praveenpuglia:
- updated browser support for `caret-color` and better description from MDN.
- updated browser support info  for `display: contents`.
- updated browser support info for logical properties `inline-start` and `inline-end` in floats.
- updated browser support info for `position: sticky`.


aeschli:
Thanks a lot, @praveenpuglia !

Issue #48, opened 2017-11-05
Remove 'Go to definition' result when it's the same as request location


tempit:
I'm writing a LS for a CSS preprocessor, and when implementing the 'Go to Definition' functionality, I see that whenever a textDocument/definition request is sent, I get back a definition with the same exact location as the original request. 
When just working with CSS this means 'Go to Definition' has no visible effect in these cases, but when using our preprocessor it means there are 2 definitions instead of one, so ctrl+click, for example, opens a definition menu instead of jumping straight to the proper place. 
For more details see: https://github.com/Microsoft/vscode/issues/37526

Can these definition location be removed when, e.g., they are the exact same as the request location?

PR #49, opened 2017-11-06
Suggested fix for issue #48

tempit:
Fix for #48 
Changed return type of findDefinition to Location[], replaced default location with empty array.


tempit:
Added replacing definition that overlaps the request location with [] (the actual original issue)

aeschli:
Thanks @tempit, I fixed it by returning null. That avoids the signature change which would be a breaking change and require a major version update.

PR #53, opened 2017-11-27
Fix Microsoft/vscode#38250. HSL2RGBA.

rebornix:
The change change simply aligns with https://github.com/Microsoft/vscode/blob/master/src/vs/base/common/color.ts#L147-L148


aeschli:
Thanks @rebornix !

PR #62, opened 2018-01-16
Emmet integration in css language service

ramya-rao-a:
Emmet completions need to be included in 2 places in a css file
- When typing a css property name
- When typing a value for a css property which expects color and what is being typed is hex color code

To keep the language service light and to avoid adding a emmet dependency (in umd) which needs to be maintained, this PR makes use of a callback model to get emmet completions.

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -13,7 +13,7 @@ import { TextDocument, Position, CompletionList, CompletionItem, CompletionItemK
 
 import * as nls from 'vscode-nls';
 const localize = nls.loadMessageBundle();
-
+const hexColorRegex = /^#[\d,a-f,A-F]+$/;
 const SnippetFormat = InsertTextFormat.Snippet;
 
 export class CSSCompletion {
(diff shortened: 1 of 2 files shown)


src/services/cssCompletion.ts
```diff
@@ -54,6 +59,9 @@ export class CSSCompletion {
 				let node = this.nodePath[i];
 				if (node instanceof nodes.Property) {
 					this.getCompletionsForDeclarationProperty(node.getParent() as nodes.Declaration, result);
+					if (this.emmetCallback) {
+						this.emmetCallback();
```
ramya-rao-a: I initially had this callback add proposals to the current result right here in the language service. Doing so would mean the cache of css completions we want to maintain in the server side would contain the emmet completions as well. This would result in looping through the whole list and removing the older emmet completions each time the emmet results had to be updated.

If the actual adding of emmet proposals to the final output is done in the server side instead, then the cache to be maintained can have just the non emmet completions, thus avoiding looping through of the list each time the emmet results have to be updated.

Thoughts?

aeschli:
Thanks @ramya-rao-a !

PR #64, opened 2018-01-19
Just a nitpick for consistency

octref:


aeschli:
Absolutely agree, the missing brackets were not intended.

PR #65, opened 2018-01-23
Handle !important as a term in scss. Fix #57

### LINKED ISSUE #57: [SCSS] Syntax Error when `!important` as arguments of Method
VSCode underlines error at `!important` syntax when it's argument of method [`if`](http://sass-lang.com/documentation/Sass/Script/Functions.html#if-instance_method):

```scss
@function get-mixin-important-var($important) {
  @return if($important, !important, null);
//                      ~~~ Error: expression expected
}
@mixin heading($important: false) {
  $i: get-mixin-important-var($important);
  font-family: "arial" $i;
}

.heading {
  @include heading(true);
}
```

However this snippet is valid when pasted to SASS Online Editor (https://www.sassmeister.com/) with option
  - Sass compiler: v3.3.14
  - Syntax: SCSS

And it's compiled to valid CSS:

```css
.heading {
  font-family: "arial" !important;
}
```

octref:


aeschli:
Looks good, thanks @octref !

PR #66, opened 2018-01-24
Allow variables as media query feature names in LESS. Fix #58

### LINKED ISSUE #58: [less] identifier expected when using @media (@var) syntax
- VSCode Version: Version 1.19.0-insider 
- OS Version: 10.11.6

Steps to Reproduce:

1. Declare a less breakpoint variable like `@m:  ~'min-width: 750px';`
2. Use it in a less file with
`@media (@m) { // rules }`

Get a [less] identifier expected error on the `@m` part.

octref:


aeschli:
Looks good, thanks @octref !

PR #67, opened 2018-02-06
Move cursor inside parenthesis for function-like values

usernamehw:
Fixes Microsoft/vscode#41018


aeschli:
@usernamehw Very nice, thanks!

Issue #69, opened 2018-02-26
parseStylesheet doesn't include the right end


ramya-rao-a:
Take this content
```
.bar {
  op0.1
}
```

At the end of `op0.1`, the offset is 14
But the cssParser parses the stylesheet and sets the `end` of the stylesheet node with 12 instead of 14

<img width="428" alt="screen shot 2018-02-25 at 10 38 45 pm" src="https://user-images.githubusercontent.com/16890566/36656321-adedecf6-1a7c-11e8-992f-d4923a347ad0.png">

The context here is that `op0.1` is a valid emmet abbreviation and emmet completions fails at this point as the offset is greater than the node(stylesheet)'s end property

PR #71, opened 2018-02-28
Update the exported type for Stylesheet

ramya-rao-a:
@aeschli 

To workaround #69, I need to know the `end` property of the stylesheet. This PR is to expose it in the exported type.

P.S: I noticed that we have already published 3.0.7 version of this module, but the `package.json` still has 3.0.6. Also the package-lock.json has a lot of changes locally when I run `npm install`. Any reason we don't want to commit those changes?

### DIFF
--- a/src/cssLanguageService.ts
+++ b/src/cssLanguageService.ts
@@ -21,7 +21,11 @@ import { SCSSCompletion } from './services/scssCompletion';
 import { LESSParser } from './parser/lessParser';
 import { LESSCompletion } from './services/lessCompletion';
 
-export type Stylesheet = {};
+export type Stylesheet = {
+	offset: number;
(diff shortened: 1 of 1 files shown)


aeschli:
Sorry, I forgot to push the commit containing the new version, done now.

As discussed on Friday, it is really a bug that the stylesheet length isn't after the `}`. This will eventually be fixed. But this will not help you with the problem that dots are not valid in property names and it's unclear what kind of node you will find at the location after a dot.

If you just want a temporary workaround for this one release, just make an `<any>` cast and access the end pos. If I make this API, it also has to behave correctly, say I have to fix the bug so that the end position is after the `}`.

ramya-rao-a:
understood

PR #73, opened 2018-03-01
Allow property completions when there is an incomplete one before it #72

### LINKED ISSUE #72: Property suggestions never show up if there is an incomplete property before
Take this content:
```
.foo  {
        margin
        m|
    }
```
There are no completions where the cursor (|) is above

ramya-rao-a:
Fixes #72


aeschli:
Oh, good catch!

PR #74, opened 2018-03-11
Fix function calls in esm mode

AviVahl:
namespace exports object is not available inside the module as `this`.


aeschli:
@AviVahl Good catch!
Just saw that there's a TS lint rule for this, I'll enable it.

PR #76, opened 2018-03-14
Add onURILiteralValue completion participation point

octref:


src/cssLanguageService.ts
```diff
@@ -54,8 +54,9 @@ export interface ColorPresentation {
 }
 
 export interface ICompletionParticipant {
-	onCssProperty: (context: { propertyName: string; range: Range; }) => void;
-	onCssPropertyValue: (context: { propertyName: string; propertyValue?: string; range: Range; }) => void;
+	onCssProperty?: (context: { propertyName: string; range: Range; }) => void;
```
aeschli: I'd write 'CSS' all uppercase (or omit it, as it's also for LESS/SCSS)
Also, I'd add interfaces for the contexts. It saves a lot of typing.

PR #79, opened 2018-03-22
Use yarn for consistency

octref:
Just to be consistent with how it's done in `/extensions`. @aeschli if you want I can do it for HTML too.


octref:
Covered by #80.

PR #80, opened 2018-03-26
CSS folding support using scanner

octref:


package.json
```diff
@@ -26,6 +26,7 @@
     "rimraf": "^2.6.2"
   },
   "dependencies": {
+    "vscode-languageserver-protocol-foldingprovider": "^1.0.1",
```
aeschli: the language service should not depend on the protocol modules as that one depends on json-rcp with 'fs' and other node dependencies.
We just need the types (FoldingRangeList, FoldingRange), which we can define ourselves in the css service
(at some point these types will go to the languageserver-types module)

src/services/cssFolding.ts
```diff
@@ -0,0 +1,83 @@
...
+		switch(token.type) {
+			case TokenType.CurlyL: {
+				stack.push(getStartLine(token));
+				break;
+			}
+			case TokenType.CurlyR: {
+				if (stack.length !== 0) {
+					const startLine = stack.pop();
+					const endLine = getEndLine(token);
+					if (startLine !== endLine) {
```
aeschli: The current folding stye is to keep showing the opening '{' and the '}' and only fold what's in between.
-> endline = getEndLine(token) - 1

src/services/cssFolding.ts
```diff
@@ -0,0 +1,83 @@
...
+						ranges.push({
+							startLine,
+							endLine,
+							type: FoldingRangeType.Region
```
aeschli: FoldingRangeType.Region means that this range is from a `#region - #endregion` range.
All normal ranges can have `undefined` as type

src/services/cssFolding.ts
```diff
@@ -0,0 +1,83 @@
...
+					}
+					break;
+				}
+			}
+			/**
+			 * In CSS, there is no single line comment prefixed with //
+			 * All comments are marked as `Comment`
+			 */
+			case TokenType.Comment: {
+				const range = tokenToRange(token, FoldingRangeType.Comment);
```
aeschli: LESS and Sass also support the `//` syntax.

Single line comments should be checked for '#region - #endregion`
All other comments must be tested whether they span more than one line.

src/test/css/folding.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+function assertRanges(lines: string[], expected: FoldingRange[]): void {
+	const document = TextDocument.create('test://foo/bar.css', 'css', 1, lines.join('\n'));
+	let actualRanges = getFoldingRegions(document).ranges;
+
+	actualRanges = actualRanges.sort((r1, r2) => r1.startLine - r2.startLine);
+	assert.deepEqual(actualRanges, expected);
+}
+
+function r(startLine: number, endLine: number, type: FoldingRangeType = FoldingRangeType.Region): FoldingRange {
```
aeschli: default type should be 'undefined'

src/test/css/folding.test.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+suite('CSS Folding - Basic', () => {
+	test('Fold single rule', () => {
+		const input = [
+			/*0*/'.foo {',
+			/*1*/'  color: red;',
+			/*2*/'}'
+		];
+		assertRanges(input, [r(0, 2)]);
```
aeschli: range should not include the closing bracket.

src/services/cssFolding.ts
octref: I cannot do this because of such test cases:

```css
.foo {
  color: red }
```

The current way it works seems fine. I saw you added column to the folding protocol. We can make use of those in the next version of css folding.

src/services/cssFolding.ts
octref: I did it but it's a confusing API. I think we should have a type for it such as `Normal`.

src/services/cssFolding.ts
aeschli: In that case the full last line stays.
It's a corner case...

src/services/cssFolding.ts
aeschli: I'm worried that adding the 'type' property everywhere would make the message that goes over the wire much bigger.

PR #82, opened 2018-04-05
Allow full expression in interpolation. Fix Microsoft/vscode#44331

octref:
This breaks a test here

```ts
testCompletionFor('@for $i from 1 through 3 { .item-#{|$i} { width: 2em * $i; } }', {
	items: [
		{ label: '$i' }
	]
});
```

However I'm wondering why we have test for this
```scss
@for $i from 1 through 3 {
  .item-#{|$i} { width: 2em * $i; }
}
```

but not these, which are far more common completion than the first one

```scss
@for $i from 1 through 3 {
  .item-#{|} { width: 2em * $i; }
}
```

```scss
@for $i from 1 through 3 {
  .item-#{$|} { width: 2em * $i; }
}
```

### DIFF
--- a/src/parser/scssParser.ts
+++ b/src/parser/scssParser.ts
@@ -165,6 +165,9 @@ export class SCSSParser extends cssParser.Parser {
 			let node = this.create(nodes.Interpolation);
 			this.consumeToken();
 			if (!node.addChild(this._parseExpr()) && !this._parseSelectorCombinator()) {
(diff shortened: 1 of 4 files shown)


aeschli:
Change looks good. But it would be interesting to figure out why the variable is no longer suggested.
Likely some other suggestions are now showing preventing the interpolation suggestions to be evaluated.

PR #84, opened 2018-04-11
Region folding. Fix Microsoft/vscode#46591

octref:


src/services/cssFolding.ts
```diff
@@ -92,6 +93,44 @@ export function getFoldingRegions(document: TextDocument): FoldingRangeList {
 			 * All comments are marked as `Comment`
 			 */
 			case TokenType.Comment: {
+				// CSS region folding
+				if (token.text === '/* #region */') {
```
aeschli: use a regexp here to allow any number of whitespaces and a description after the 'region'.

src/test/css/folding.test.ts
```diff
@@ -180,6 +180,19 @@ suite('CSS Folding - Nested', () => {
...
+			/*0*/'/* #region */',
+			/*1*/'& .bar {',
+			/*2*/'  color: red;',
+			/*3*/'}',
+			/*4*/'/* #endregion */'
+		];
+		assertRanges(input, [r(0, 4, 'region'), r(1, 2)]);
+	});
+});
+
```
aeschli: please add some tests with incorrect nesting, e.g. the
```
/* #region */
& .bar {
/* #endregion  */
}
```
The region should win...

aeschli: Regions can not intersect. Maybe look at this example:
```
/* #region */
.bar {
/* #endregion  */
// hello
}
```
[0,2] and [1,3] would be illegal.
-> once you see the /* #endregion  */, discard all open ranges until you see the open /* #region */.

octref: OK, will open a new issue to track this.

aeschli: I created #85.

PR #89, opened 2018-04-12
maxranges, fix #88

### LINKED ISSUE #88: [folding] limit number of ranges created
The LSP folding API provides a number of max ranges the that the server should create.
We should pass that limit to the service and limit the number of ranges

octref:


src/services/cssFolding.ts
```diff
@@ -79,11 +79,24 @@ export function getFoldingRegions(document: TextDocument): FoldingRangeList {
 					}
 
 					if (startLine !== endLine) {
-						ranges.push({
-							startLine,
-							endLine,
-							type: undefined
-						});
+						if (maxRanges) {
```
aeschli: Of course we can just cut the last regions, but that is not the best user experience.
In HTML and JSON implementatio I filter out the most nested ranges.

src/cssLanguageService.ts
```diff
@@ -94,7 +94,7 @@ export interface LanguageService {
 	findDocumentColors(document: TextDocument, stylesheet: Stylesheet): ColorInformation[];
 	getColorPresentations(document: TextDocument, stylesheet: Stylesheet, color: Color, range: Range): ColorPresentation[];
 	doRename(document: TextDocument, position: Position, newName: string, stylesheet: Stylesheet): WorkspaceEdit;
-	findFoldingRegions(document: TextDocument, stylesheet: Stylesheet): FoldingRangeList;
+	findFoldingRegions(document: TextDocument, maxRanges?: number | undefined): FoldingRangeList;
```
aeschli: Note that this is a breaking API change and will require an increase of a major version update.

aeschli: Oh, we are on a next release. No problem then!

aeschli:
merged, we can implement a better filter strategy later

src/services/cssFolding.ts
octref: Extracted to #90.

Issue #68, opened 2018-02-08
Use MDN data to power the CSS language server


auchenberg:
I just had a chat with the MDN team, and they have published a NPM module that enables tools to take advantage of the detailed MDN data.

We should explore using https://github.com/mdn/browser-compat-data to power this language server

praveenpuglia:
This is something we had a discussion on but no progress was made. 
https://github.com/Microsoft/vscode-css-languageservice/issues/40

octref:
Some more insights:

- mdn/data has data for syntax of css properties, but not descriptions
- Using MDN data could make us always up-to-date on newly added properties. An example: for this one https://github.com/Microsoft/vscode/issues/46423, the [mdn data](https://github.com/mdn/data/blob/31e1b207c71ad1cc6f5013875740ff74cf5a9acb/css/at-rules.json) is:

  ```json
      "font-display": {
        "syntax": "[ auto | block | swap | fallback | optional ]",
        "media": "visual",
        "percentages": "no",
        "initial": "auto",
        "computed": "asSpecified",
        "order": "uniqueOrder",
        "status": "experimental"
      }
  ```

octref:
An analysis on CSS properties:

- Our side: Generated [`browsers.ts`](https://github.com/Microsoft/vscode-css-languageservice/blob/master/src/data/browsers.ts)
- MDN: [at-rules](https://github.com/mdn/data/blob/master/css/at-rules.json) and [properties](https://github.com/mdn/data/blob/master/css/properties.json)

When we compare our browser data with MDN data, there are 239 rules in our side and not in MDN side, here is a complete list:

<details>
  <summary>Properties in VSC not in MDN</summary>

  ```
  alt
  behavior
  clip-rule
  color-interpolation-filters
  enable-background
  fill
  fill-opacity
  fill-rule
  flood-color
  flood-opacity
  glyph-orientation-horizontal
  glyph-orientation-vertical
  kerning
  lighting-color
  marker
  marker-end
  marker-mid
  marker-start
  motion
  motion-offset
  motion-path
  motion-rotation
  -moz-animation
  -moz-animation-delay
  -moz-animation-direction
  -moz-animation-duration
  -moz-animation-iteration-count
  -moz-animation-name
  -moz-animation-play-state
  -moz-animation-timing-function
  -moz-backface-visibility
  -moz-background-clip
  -moz-background-inline-policy
  -moz-background-origin
  -moz-border-image
  -moz-box-align
[...]

PR #91, opened 2018-04-16
[WIP] Use MDN to enhance CSS LS. Part of #68

octref:
This PR:

- Adds dependency on `mdn-data` (can we switch to yarn.lock instead of package-lock.json entirely?)
- Generate `browsers.ts` by combining VSC and MDN data, while giving preference to VSC data
- Show `syntax` and `status` in completion item's documentation.

Questions:

- How should we generate restrictions / values? Currently for any of the new properties that MDN data introduces, we have no restriction and the value auto completion is inaccurate.
- Our data contains [239 properties](https://github.com/Microsoft/vscode-css-languageservice/issues/68#issuecomment-380172011) that's not in MDN Data. However, a majority of them are actually SVG properties (tracked in https://github.com/mdn/data/issues/59). We can either
  - Hand curate the 239 properties to tell which ones are not SVG related, and mark those as deprecated / obsolete.
  - Wait for https://github.com/mdn/data/issues/59


aeschli:
Note that`browsers.js` should be as small as possible. vscode-css-languageservice is also used in the browser (for monaco) and we want the minimal memory footprint as well as download package size.
So `browsers.js` should only keep the properties that are used in the code.

aeschli:
I like the use of status. Maybe we should omit `"status": "standard"` in order to make browser.js smaller.

octref:
> Maybe we should omit "status": "standard" in order to make browser.js smaller.

@aeschli Sure, I can remove it together with a bunch of other unused properties.

@chriseppstein Yes, I submitted license review request for mdn/data. Before the review passes we won't ship the change. Meanwhile, can you chime in at https://github.com/mdn/data/issues/199 for your concerns?

octref:
@chriseppstein 

https://www.mozilla.org/en-US/MPL/2.0/FAQ/

> Q11: How 'viral' is the MPL? If I use MPL-licensed code in my proprietary application, will I have to give all the source code away?
No. The license requires that Modifications (as defined in Section 1.10 of the license) must be licensed under the MPL and made available to anyone to whom you distribute the Source Code. However, new files containing no MPL-licensed code are not Modifications, and therefore do not need to be distributed under the terms of the MPL, even if you create a Larger Work (as defined in Section 1.7) by using, compiling, or distributing the non-MPL files together with MPL-licensed files. This allows, for example, programs using MPL-licensed code to be statically linked to and distributed as part of a larger proprietary piece of software, which would not generally be possible under the terms of stronger copyleft licenses.

We don't do any modification to the original mdn/data source, so vscode-css-languageservice could still be shared under MIT license.

chriseppstein:
So, the legal definition of modification isn’t the same as a developer’s definition. I haven’t reviewed this PR in depth yet, but according to the lawyers I’ve worked with there’s a number of build time transformations that meet the legal definition of “modification” and would require a dual license.

If you don’t do anything like that, it’s still something that will require an ongoing attention towards.

octref:
@chriseppstein Thanks, I've opened https://github.com/mdn/data/issues/210

PR #96, opened 2018-05-22
[css] Parse grid-template value correctly

octref:


src/parser/cssParser.ts
```diff
@@ -422,15 +422,23 @@ export class Parser {
...
+		const declarationName = this.prevToken.text;
 
 		if (!this.accept(TokenType.Colon)) {
 			return <nodes.Declaration>this.finish(node, ParseError.ColonExpected, [TokenType.Colon], resyncStopTokens);
 		}
 		node.colonPosition = this.prevToken.offset;
 
-		if (!node.setValue(this._parseExpr())) {
-			return this.finish(node, ParseError.PropertyValueExpected);
+		if (declarationName === 'grid-template') {
```
aeschli: Make the parser depending on the property name is problematic. The property name could be a variable name (--foo). In SCSS/LESS property names can contain the parent selector (&) or interpolation, so it is hard to know what the actual property name is.
I'd rather make this a valid value for all declarations.

aeschli: I'd suggest to accept this._parseNamedLine() as a term:
See my suggested fix: https://github.com/Microsoft/vscode-css-languageservice/tree/aeschli/grid-template

octref: I see -- that's much cleaner. Thanks!
I also put a reference https://www.w3.org/TR/CSS21/grammar.html in the comment.

PR #97, opened 2018-05-23
[less] Handle Ruleset passed into detached ruleset reference as argument. Fix Microsoft/vscode#43333

octref:


src/parser/lessParser.ts
```diff
@@ -146,7 +146,7 @@ export class LESSParser extends cssParser.Parser {
 	}
 
 	public _parseDetachedRuleSetBody(): nodes.Node {
-		return this._tryParseKeyframeSelector() || super._parseRuleSetDeclaration();
+		return this._tryParseKeyframeSelector() || super._parseRuleSetDeclaration() || super._parseRuleset();
```
aeschli: Should be:
`this._tryParseKeyframeSelector() || this._tryParseRuleset(true)  || super._parseRuleSetDeclaration()`
Rule sets start with selectors that also contains ':'. _tryParseRuleset will try to parse this a selector, but will reset the scanner if it turns out to be something else.

octref:
I rebased on the PR for grid-template - since a change there is affecting one of the tests. So this can be merged after #96.

PR #100, opened 2018-05-23
Switch to yarn. Fix #98

### LINKED ISSUE #98: Switch to yarn?
I see 966d558743fbdb6db5f99d605cca0976f92af6a1 and 8dcf82c79a267e4d02febd56afc63f3d1dd66af0. We have both `package-lock.json` and `yarn.lock` in the repo.

Can we stick to one of them? Since we are using `yarn` in VS Code I suggest we stick with it.

octref:
- Travis already uses `yarn` when `yarn.lock` is found:
  image
- Adds caching for `yarn`. The cache is global at `$HOME/.yarn-cache` and each time `yarn install` will either pull cached dependencies from it, or install new ones


connorshea:
Should the README.md be updated to use `yarn install`? https://github.com/Microsoft/vscode-css-languageservice#installation

PR #101, opened 2018-05-30
Handle at rules before cursor in declaration completions. Fix Microsoft/vscode#50805

octref:


src/services/scssCompletion.ts
```diff
@@ -211,6 +211,20 @@ export class SCSSCompletion extends CSSCompletion {
 		return result;
 	}
 
+	public getCompletionsForDeclarations(declarations: nodes.Declarations, result: CompletionList): CompletionList {
+		if (!declarations || this.offset === declarations.offset) { // incomplete nodes
+			return result;
+		}
+
+		let node = declarations.findFirstChildBeforeOffset(this.offset);
```
aeschli: I think we should add that directly to cssCompletion.getCompletionsForDeclarations.

aeschli:
I pushed my review comments in order to publish a new version of the vscode-css-languageservice that we can consume in vscode.

PR #104, opened 2018-06-06
Update display of supported browsers for CSS properties on-hover.

connorshea:
I wanted to try messing around with extension development, but more importantly I wanted it clear what the browsers list is supposed to mean.

Old style: '(Chrome 31, Firefox 27, IE 11, Edge)'
New style: 'Browser Support: Chrome 31, Firefox 27, IE 11, Edge'

Before:

screen shot 2018-06-06 at 4 14 47 pm

After:

screen shot 2018-06-06 at 4 14 19 pm

### DIFF
--- a/src/services/languageFacts.ts
+++ b/src/services/languageFacts.ts
@@ -595,7 +595,7 @@ export function getEntryDescription(entry: { description: string; browsers: Brow
 
 	let browserLabel = getBrowserLabel(entry.browsers);
 	if (browserLabel) {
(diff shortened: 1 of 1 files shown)


octref:
For me I think it's actually too much noise. For all properties that's `standard` and implemented across all browsers we should just hide the browser support part.

aeschli:
IMO its pretty clear what the browser list means. I'd rather not make a change.

connorshea:
That’s fine by me 👍 you can close this PR then :)

PR #116, opened 2018-08-21
Add CSS Lint Rule name to lint markers.

connorshea:
Resolves https://github.com/Microsoft/vscode/issues/56925

Before:
screen shot 2018-08-21 at 2 16 06 pm
screen shot 2018-08-21 at 2 16 15 pm

After:
screen shot 2018-08-21 at 2 10 55 pm
screen shot 2018-08-21 at 2 11 09 pm

The rule name doesn't include the `css.lint.*` prefix, but that isn't included in the Rule object that `marker.getRule()` returns. That can be worked around using `document.languageId` but I'm fairly sure that'd break on some edge cases.

### DIFF
--- a/src/services/cssValidation.ts
+++ b/src/services/cssValidation.ts
@@ -35,7 +35,7 @@ export class CSSValidation {
 			return <Diagnostic>{
 				code: marker.getRule().id,
 				source: document.languageId,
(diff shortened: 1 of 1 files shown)


aeschli:
Thanks @connorshea !

PR #118, opened 2018-09-18
Handle dash-number interpolations in SCSS. Fix Microsoft/vscode#56743

octref:


aeschli:
Fix is good, thanks @octref !

PR #119, opened 2018-09-18
Handle trailing commas in SCSS. Fix Microsoft/vscode#57858

octref:


aeschli:
Thanks for the fix! 

I added some changes to to use 'break' the while loop and use the normal code to handle the rest of the statement.

PR #120, opened 2018-09-19
Make document highlight mark whole className. Fix Microsoft/vscode#56354

octref:


src/parser/cssParser.ts
```diff
@@ -1198,9 +1198,12 @@ export class Parser {
 			return null;
 		}
 		let node = this.createNode(nodes.NodeType.ClassSelector);
-		this.consumeToken(); // '.'
-		if (this.hasWhitespace() || !node.addChild(this._parseSelectorIdent())) {
-			return this.finish(node, ParseError.IdentifierExpected);
+		if (this.acceptDelim('.')) {
```
aeschli: don't think the `else` clause can be reached. there's a test for `.` just a few lines above.

aeschli:
Wow, that's a surprisingly easy fix. Thanks @octref !

src/parser/cssParser.ts
octref: I see. I was trying to follow `_parseHash`, but just realized that `!this.peek(TokenType.Hash) && !this.peekDelim('#')` covers more case than `this.acceptDelim('#')`, thus the extra if-else check.

PR #131, opened 2018-10-30
Rename specificity label and add a link to MDN

tlent:
Fixes https://github.com/Microsoft/vscode/issues/62155

I renamed it "Selector specificity" instead of "CSS specificity" because the user would already know they are editing a CSS file and "Selector specificity" seemed like a better description.

I wasn't sure if the link should be included in the `localize()` call or not. MDN has that page in other languages so maybe it should be localized to link to the appropriate language page? If this is not how it should be done I can change it.

This is my first VS Code related contribution. I hope that it is helpful.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-  "editor.insertSpaces": false
+  "editor.insertSpaces": false,
+  "prettier.useTabs": true,
+  "prettier.printWidth": 120,
+  "prettier.semi": true
 }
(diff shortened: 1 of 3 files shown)


octref:
I made some changes and pulled it in. Mainly because of the merge conflicts.

Thanks again for the PR and don't forget to get your Microsoft shirt for Hacktoberfest! Today is the last day. https://open.microsoft.com/2018/09/30/join-hacktoberfest-2018-celebration-microsoft/

PR #135, opened 2018-12-02
Support for Less root functions, lookups, anonymous mixins

matthew-dean:
This fixes a number of errors in the language service parser for Less.

- In Less, functions can be called anywhere to create a node, including the root.
- Custom functions can accept mixin-like arguments (such as accepting rulesets), and arguments can be separated with a semi-colon
- Adds support for map lookups `@ruleset[@lookup]`
- Adds (partial) support for property references `$prop` - still needs interpolation support
- Adds support for anonymous mixins, introduced with the `each()` function

There's some functions to add, like `if()`, `range()`, `each()` and some other things, but this should give much fewer errors for people using Less in VSCode.

### DIFF
--- a/src/parser/lessParser.ts
+++ b/src/parser/lessParser.ts
@@ -29,6 +29,7 @@ export class LESSParser extends cssParser.Parser {
 
 		return this._tryParseMixinDeclaration()
 			|| this._tryParseMixinReference(true)
+			|| this._parseFunction()
 			|| this._parseRuleset(true);
 	}
 
@@ -115,7 +116,7 @@ export class LESSParser extends cssParser.Parser {
 		let node = <nodes.VariableDeclaration>this.create(nodes.VariableDeclaration);
 
 		let mark = this.mark();
-		if (!node.setVariable(this._parseVariable())) {
+		if (!node.setVariable(this._parseVariable(true))) {
 			return null;
 		}
 
@@ -139,11 +140,36 @@ export class LESSParser extends cssParser.Parser {
 	}
 
 	public _parseDetachedRuleSet(): nodes.Node {
+		let mark = this.mark();
+
+		// "Anonymous mixin" used in each() and possibly a generic type in the future
+		if (this.peekDelim('#') || this.peekDelim('.')) {
+			this.consumeToken();
+			if (!this.hasWhitespace() && this.accept(TokenType.ParenthesisL)) {
+				let node = <nodes.MixinDeclaration>this.create(nodes.MixinDeclaration);
+
+				if (node.getParameters().addChild(this._parseMixinParameter())) {
+					while (this.accept(TokenType.Comma) || this.accept(TokenType.SemiColon)) {
+						if (this.peek(TokenType.ParenthesisR)) {
+							break;
+						}
+						if (!node.getParameters().addChild(this._parseMixinParameter())) {
+							this.markError(node, ParseError.IdentifierExpected, [], [TokenType.ParenthesisR]);
+						}
+					}
+				}
+		
+				if (!this.accept(TokenType.ParenthesisR)) {
+					this.restoreAtMark(mark);
+					return null;
+				}
+			}
+		}
+
 		if (!this.peek(TokenType.CurlyL)) {
 			return null;
 		}
 		let content = <nodes.BodyDeclaration>this.create(nodes.BodyDeclaration);
-
 
 		this._parseBody(content, this._parseDetachedRuleSetBody.bind(this));
 		return this.finish(content);
@@ -153,22 +179,65 @@ export class LESSParser extends cssParser.Parser {
 		return this._tryParseKeyframeSelector() || this._tryParseRuleset(true) || super._parseRuleSetDeclaration();
 	}
 
-	public _parseVariable(): nodes.Variable {
-		if (!this.peekDelim('@') && !this.peek(TokenType.AtKeyword)) {
+	public _addLookupChildren(node: nodes.Node): boolean {
+		if (!node.addChild(this._parseLookupValue())) {
+			return false;
+		}
+
+		let expectsValue = false;
+
+		while (true) {
+			if (this.peek(TokenType.BracketL)) {
+				expectsValue = true;
+			}
+			if (!node.addChild(this._parseLookupValue())) {
+				break;
+			}
+			expectsValue = false;
+		}
+		return !expectsValue;
+	}
+
+	public _parseLookupValue(): nodes.Node {
+		let node = <nodes.Node>this.create(nodes.Node);
+		let mark = this.mark();
+		
+		if (!this.accept(TokenType.BracketL)) {
+			this.restoreAtMark(mark);
+			return null;
+		}
+		if ((this.accept(TokenType.Ident) || this.accept(TokenType.Num) || this._parseVariable(false, true)) && this.accept(TokenType.BracketR)) {
+			return <nodes.Node>node;
+		}
+		this.restoreAtMark(mark);
+		return null;
+	}
+
+	public _parseVariable(declaration: boolean = false, insideLookup: boolean = false): nodes.Variable {
+		const isPropertyReference = !declaration && this.peekDelim('$');
+		if (!this.peekDelim('@') && !isPropertyReference && !this.peek(TokenType.AtKeyword)) {
 			return null;
 		}
 
 		let node = <nodes.Variable>this.create(nodes.Variable);
 		let mark = this.mark();
-		while (this.acceptDelim('@')) {
+		
+		while (this.acceptDelim('@') || (!declaration && this.acceptDelim('$'))) {
 			if (this.hasWhitespace()) {
 				this.restoreAtMark(mark);
 				return null;
 			}
 		}
-		if (!this.accept(TokenType.AtKeyword)) {
+
+		if (!this.accept(TokenType.AtKeyword) && !this.accept(TokenType.Ident)) {
 			this.restoreAtMark(mark);
 			return null;
+		}
+		if (!insideLookup && this.peek(TokenType.BracketL)) {
+			if (!this._addLookupChildren(node)) {
+				this.restoreAtMark(mark);
+				return null;	
+			}
 		}
 		return <nodes.Variable>node;
 	}
@@ -253,6 +322,7 @@ export class LESSParser extends cssParser.Parser {
 		return this._tryParseMixinDeclaration()
 			|| this._tryParseRuleset(true)  // nested ruleset
 			|| this._tryParseMixinReference() // less mixin reference
+			|| this._parseFunction()
 			|| this._parseExtend() // less extend declaration
 			|| super._parseRuleSetDeclaration(); // try css ruleset declaration as the last option
 	}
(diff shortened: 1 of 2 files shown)


matthew-dean:
Just to note: I tested this through just the regular (and updated) tests on this repo. I haven't tried this in a local development build of VSCode.

matthew-dean:
One thing that's still kind of outstanding is that because rulesets can be maps in Less, the warning and squiggly lines about "unknown property"s is somewhat useless. Is there a way for those `unknown property` flags to not be set in a Less document? I think just coloring the keyword differently would be enough, but the squiggly line is quite obnoxious. (That would be true for CSS in general IMO. It's okay to visually distinguish known and unknown, but I wouldn't hit the user over the head.)

matthew-dean:
...Ping?

matthew-dean:
@octref Ok no worries, just wanted to make sure it didn't get lost!

src/test/less/parser.test.ts
```diff
@@ -123,7 +144,13 @@ suite('LESS - Parser', () => {
 		let parser = new LESSParser();
 		assertNode('%()', parser, parser._parseFunction.bind(parser));
 		assertNoNode('% ()', parser, parser._parseFunction.bind(parser));
+		
+		assertNode('func(a, b; bar)', parser, parser._parseRuleSetDeclaration.bind(parser));
+		assertNode('func({a: b();}, bar)', parser, parser._parseRuleSetDeclaration.bind(parser));
+		assertNode('func(.(@val) {})', parser, parser._parseRuleSetDeclaration.bind(parser));
```
octref: Is this allowed?

This in lessc 3.9 gives me errors:

```less
func(@x) {
  @x();
}

.foo {
  func({a: b();});
}
```

```
 less-test > lessc test.less
ParseError: Unrecognised input in /Users/octref/Code/work/less-test/test.less on line 12, column 10:
11 
12 func(@x) {
13   @x();
```

I thought functions only start with dot, like `.func() {}`?

src/test/less/parser.test.ts
```diff
@@ -252,11 +285,15 @@ suite('LESS - Parser', () => {
...
 		assertNode('.foo { @supports(display: grid) { display: none; }}', parser, parser._parseRuleset.bind(parser));
+		assertNode('.parent { color:green; @document url-prefix() { .child { color:red; }}}', parser, parser._parseStylesheet.bind(parser));
+		assertNode('@supports (property: value) { .outOfMedia & { @media (max-size: 2px) { @supports (whatever: something) { property: value;}}}}', parser, parser._parseStylesheet.bind(parser));
+		
 	});
 
 	test('Interpolation', function () {
 		let parser = new LESSParser();
 		assertNode('.@{name} { }', parser, parser._parseRuleset.bind(parser));
+		assertNode('.${name} { }', parser, parser._parseRuleset.bind(parser));
```
octref: I believe this is "Adds support for property references ($prop and interpolated ${prop})" you mentioned, but I couldn't find any such example on less doc. Can you give an example?

src/parser/lessParser.ts
```diff
@@ -313,29 +394,69 @@ export class LESSParser extends cssParser.Parser {
...
 	private peekInterpolatedIdent() {
-		return this.peek(TokenType.Ident) || this.peekDelim('@') || this.peekDelim('-');
+		return this.peek(TokenType.Ident) ||
+			this.peekDelim('@') ||
+			this.peekDelim('$') ||
+			this.peekDelim('-');
 	}
 
-	public _acceptInterpolatedIdent(node: nodes.Node): boolean {
+	public acceptRegexp(regEx: RegExp): boolean {
```
octref: Maybe move this to cssparser.

src/parser/lessParser.ts
```diff
@@ -313,29 +394,69 @@ export class LESSParser extends cssParser.Parser {
...
+		if (regEx.test(this.token.text)) {
+			this.consumeToken();
+			return true;
+		}
+		return false;
+	}
+
+	public _parseRegexp(regEx: RegExp): nodes.Node {
```
octref: Same as above, seems useful enough that we can have it in cssparser.

src/parser/lessParser.ts
```diff
@@ -348,7 +469,13 @@ export class LESSParser extends cssParser.Parser {
...
 			return null;
 		};
-		while (this.accept(TokenType.Ident) || node.addChild(this._parseInterpolation() || this.try(delimWithInterpolation))) {
+		const accept = identRegex ?
+			(): boolean => this.acceptRegexp(identRegex) :
+			(): boolean => this.accept(TokenType.Ident);
+
+		while (accept() || 
+			node.addChild(this._parseInterpolation() || 
+			this.try(delimWithInterpolation))) {
```
octref: You can the conditions in the same line. It's hard to read what's inside the `while` because of no indentation.

src/parser/lessParser.ts
```diff
@@ -361,9 +488,10 @@ export class LESSParser extends cssParser.Parser {
 	}
 
 	public _parseInterpolation(): nodes.Node {
-		//  @{name}
+		// @{name} Variable or 
+		// ${name} Property
```
octref: Same as the other `${prop}` confusion. Can you point me to the docs that describe this feature?

src/test/less/parser.test.ts
```diff
@@ -16,11 +16,23 @@ suite('LESS - Parser', () => {
 	test('Variable', function () {
 		let parser = new LESSParser();
 		assertNode('@color', parser, parser._parseVariable.bind(parser));
+		assertNode('$color', parser, parser._parseVariable.bind(parser));
+		assertNode('$$color', parser, parser._parseVariable.bind(parser));
+		assertNode('@$color', parser, parser._parseVariable.bind(parser));
+		assertNode('$@color', parser, parser._parseVariable.bind(parser));
```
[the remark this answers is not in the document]
matthew-dean: This is a combination of two things.

1. Less supports variables and (like PHP) variable variables. Meaning, the name (identifier) of a variable being referenced can itself be variable.  So: `@@var` means "return the value of a variable with the id that `@var` resolves to. If `@var` is `foo` then `@foo` is looked up/returned.
2. In a 2.x release, support for property referencing (`$prop`) was added. Basically it just means that properties are (more or less) variables.

Well it's not explicitly documented, you can technically then combine these features of variable variables, where the property identifier being looked up is variable, or the variable id being looked up is variable based on a property name.

Why someone would ever do this, I don't know, it's just a normal side-effect of combining these features.

src/test/less/parser.test.ts
matthew-dean: You're thinking of mixins. Mixins start with `.` or `#`, and those are defined in your stylesheet. Functions are defined in JavaScript. They accept a node (or nodes), and return a node. `lighten()` and `darken()` are examples of Less functions. After Less made it easier for users to define functions using `@plugin`, the restriction on functions being only in a property's value was dropped. This way, authors can call a function that returns a ruleset, for example. Less also later added functions that are meant to be called around rules, such as `each()`.

src/test/less/parser.test.ts
matthew-dean: Similar to what's mentioned above, this is just a normal side-effect of treating properties as variables. That is, you can do interpolation in other values. Would this particular syntax in the above assertion ever make sense? Not really. But it's valid Less, which should be what the parser is checking for.

src/parser/lessParser.ts
matthew-dean: I could probably do that.

src/parser/lessParser.ts
matthew-dean: Fair enough. Or maybe for really long conditions, move them all to an indented form?

As in:

```js
while(
  condition() ||
  condition2()
) {
}
```
?

src/parser/lessParser.ts
matthew-dean: The documentation is very brief and doesn't cover the interpolated form. It's here: http://lesscss.org/features/#variables-feature-properties-as-variables-new-

It should probably be added to Less docs, but it's valid. In a lot of cases for these, I pulled code directly from Less unit tests and added them to the language service tests. In fact, I opened Less unit tests in VSCode to see what was failing first, before adding language service tests.

src/test/less/parser.test.ts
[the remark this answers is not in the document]
octref: This doesn't compile with less 3.9:

```lessc
@color: #fff;

@lookup: 'color';

.foo {
  color: @@lookup;
  background-color: $@lookup;
}
```

I guess we can add more less-specific language features in the future. Now as long as VS Code doesn't fail at parsing these less files, users can get by.

src/parser/lessParser.ts
octref: Yeah that's better. `this.try` and `hasContent` on the same indent level was confusing to me.

src/test/less/parser.test.ts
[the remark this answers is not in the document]
matthew-dean: @octref You're right that that usage may not compile, but this does:
```less
@dr: {
  prop: value;
}

.box {
  @lookup: prop;
  foo: @dr[$@lookup];
}
```
For lookup values, I think I did the VSCode parsing to use any of the variable reference form. So, yeah, it may not compile, but VSCode should still not throw an unrecoverable parsing error.

It's also possible the code you wrote will work at some point, just for consistency.

aeschli:
@matthew-dean Thanks a lot for your help! We're happy to have your expertise!

Most of the changes are good. To (hopefully) make it simpler to work on this, I already pushed some of the uncontroversial changes, see 5e7a590babf332ce4a290276805aaaef12b14334

What I don't like so much are changes related to the map lookups. The lookup is modelled as a child to variables and mixins. I think it is cleaner if we do it in the term, like an an operator (e.g. the % operator).

matthew-dean:
@aeschli 

> What I don't like so much are changes related to the map lookups. The lookup is modelled as a child to variables and mixins. I think it is cleaner if we do it in the term, like an an operator (e.g. the % operator).

I'm unclear what you're saying...?

aeschli:
@matthew-dean My suggestion would be to add the support to `parseTerm`. Something like:
```
	public _parseTerm(): nodes.Term {
		let term = this._parseSimpleTerm();
		if (term && this.accept(TokenType.BracketL)) {
			// TODO: check for name, index of variable maybe even just `term`?
			if (!this.accept(TokenType.BracketR)) {
				//return error ...
			}
		}
		return term;
	}


	public _parseSimpleTerm(): nodes.Term {
		let term = super._parseTerm();
		if (term) { return term; }

		term = <nodes.Term>this.create(nodes.Term);
		if (term.setExpression(this._parseVariable()) ||
			term.setExpression(this._parseEscaped()) ||
			term.setExpression(this._tryParseMixinReference())) {

			return <nodes.Term>this.finish(term);
		}

		return null;
	}
```

That allows indexed access of every term which is probably more than necessary, but would keep the variables and mixins as is.

matthew-dean:
Sorry, will try to get back to this soon and the requested changes when I can find some time.

radium-v:
I would love to see this moved further through the pipeline. Broken map lookups cause every proceeding colon, bracket, and curly brace to be presented as an error. It's really frustrating.

image

matthew-dean:
@radium-v Would you have time to fix conflicts and address the few changes requested? I've just got a full schedule at work + parenting.

radium-v:
@matthew-dean sure, I'll dive in this week.

kevinramharak:
Sooo, is anyone still on this? As it has missed its milestones and I would love to see this implemented but do not have the skills to finish this.

matthew-dean:
@kevinramharak I apologize. As someone who coordinates open source work, I know it's frustrating when a PR seems so close but languishes in changes. I just haven't had the extra cycles for any development on the side. Hopefully @radium-v can get to it? On a long enough timeline, I should be able to return to a pile of Less maintenance stuff, which includes this, but I don't yet have a slot in my schedule.

radium-v:
@kevinramharak @matthew-dean I also apologize, I overpromised when I volunteered to pick this up. I do still intend to pick it up and run it to the finish line, This is a blocker for radium-v/better-less#14 and I'm finally able to get back to it.

matthew-dean:
@octref @aeschli 

I've merged master and moved methods according to feedback. However, I found that with even the base-level merge I was getting unrelated errors on SCSS (which was logging console errors but not reporting as errors to the test runner?) and an error with a property in the CSS actions test.

Can you take a look?

octref:
It's too late in our development week, but I'll try to review and make sure this goes in for August. Thanks again for your work!

matthew-dean:
@octref No problem!

PR #137, opened 2018-12-10
Don't warn for unknown properties inside an :export block

### LINKED ISSUE #103: scss: Unknown property when using :export, CSS Modules
Hello!

I'm using a fairly normal Webpack + PostCSS + CSS Modules setup, and one of the features of CSS Modules is that any properties under the tag `:export` are exported as part of the imported Webpack file.

Example:

```scss
:export {
  myColor: red;
}
```

```javascript
import styles from . . .

styles.myColor; // red
```

Currently I get an error message `Unknown property: 'myColor'` which makes sense as its not in the CSS or Sass specs, but I don't want to have to disable the entirety of CSS Language Features to not see these warnings;

The same thing happens for the `composes` property of CSS Modules, which you can use like so . . .

```css
.a {
  background-color: red;
}

.b {
  composes: a;
  color: white;
}
```

This feature is less important since nearly all CSS preprocessors support some sort of inheritance.

Is better integration with CSS Modules something you'd be willing to support? Perhaps these features could only be enabled if the file ended with `.mod.css`, `.mod.scss`, etc.

Thank you!

kristerkari:
Currently custom properties inside a CSS modules `:export` block are reported as unknown. The properties inside `:export` are not CSS properties - they are custom properties that are exported from the CSS/SCSS file to Javascript (e.g. variables).

This PR adds a simple check to see if the selector is `:export`, and skips the whole checking for unknown properties in that case.

Please enable `Diff settings > hide whitespace changes` from Github's UI when reviewing the PR.

fixes #103

### DIFF
--- a/src/services/lint.ts
+++ b/src/services/lint.ts
@@ -411,66 +411,70 @@ export class LintVisitor implements nodes.IVisitor {
 		//	Unknown propery & When using a vendor-prefixed gradient, make sure to use them all.
 		/////////////////////////////////////////////////////////////
 
-		let propertiesBySuffix = new NodesByRootMap();
-		let containsUnknowns = false;
+		let isExportBlock = node.getSelectors().getText() === ":export";
 
-		for (let node of declarations.getChildren()) {
-			if (this.isCSSDeclaration(node)) {
-				let decl = <nodes.Declaration>node;
-				let name = decl.getFullPropertyName().toLowerCase();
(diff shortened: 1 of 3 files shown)


octref:
Sorry both of us was on vacation. I'll review it in December.

kristerkari:
Thanks! Have nice holidays

octref:
LGTM. Thanks for your contribution!

PR #138, opened 2018-12-27
Improve box model analysis

yume-chan:
I always have `css.lint.boxModel` on, but what I usually want to do is to override properties from other css files.

``` css
.element {
    width: 60px;
    padding: 0 !important;
}
```

The problem is, VS Code keeps warning me about this: **Do not use width or height when using padding or border.**

Does `padding: 0` really counts as *using*?

So instead of turn this feature off, I'd love to make it be aware of what I'm doing.

### DIFF
--- a/src/services/boxModel.ts
+++ b/src/services/boxModel.ts
@@ -1,0 +1,234 @@
+
+import * as nodes from '../parser/cssNodes';
+import { includes } from '../utils/arrays';
+
+export class Element {
+
+	public name: string;
+	public node: nodes.Declaration;
+
+	constructor(text: string, data: nodes.Declaration) {
+		this.name = text;
+		this.node = data;
+	}
+}
+
+interface SideState {
+	value: boolean;
+
+	properties: Element[];
+}
+
+interface BoxModel {
+	width?: Element;
+
+	height?: Element;
+
+	top: SideState;
+
+	right: SideState;
+
+	bottom: SideState;
+
+	left: SideState;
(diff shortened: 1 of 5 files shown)


src/services/boxModel.ts
```diff
@@ -0,0 +1,234 @@
...
+	for (const child of children) {
+		const value = child.getText();
+		if (!checkLineWidth(value) || !checkLineStyle(value, /* allowsKeywords: */ false)) {
+			return false;
+		}
+	}
+	return true;
+}
+
+export default function calaculateBoxModel(propertyTable: Element[]): BoxModel {
```
KamasamaK: Should be `calculateBoxModel`

src/utils/arrays.ts
```diff
@@ -24,3 +24,19 @@ export function findFirst<T>(array: T[], p: (x: T) => boolean): number {
 	}
 	return low;
 }
+
+export function includes<T>(array: T[], item: T): boolean {
```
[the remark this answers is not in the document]
yume-chan: I can't find any statement about engine requirement (which version of node or browser it will run on), so I didn't modify `lib` config.

src/services/boxModel.ts
yume-chan: Fixed.

aeschli:
Thanks @KamasamaK, I reviewed the changes, but renamed the new file to be also used by future lint rules.

PR #139, opened 2018-12-31
[fix] calculate selector-specificity from root node - fix Microsoft/vscode#63200

ShPelles:
[fix] calculate selector-specificity from root node - fix Microsoft/vscode#63200

### DIFF
--- a/src/services/selectorPrinting.ts
+++ b/src/services/selectorPrinting.ts
@@ -358,7 +358,14 @@ function selectorToSpecificityMarkedString(node: nodes.Node): MarkedString {
 export function selectorToMarkedString(node: nodes.Selector): MarkedString[] {
 	let root = selectorToElement(node);
 	let markedStrings = new MarkedStringPrinter('"').print(root);
-	markedStrings.push(selectorToSpecificityMarkedString(node));
+
+	let parentRuleSet = getParentRuleSets(node).pop();
+	let rootNode: nodes.Node = node;
+	if (!!parentRuleSet) {
+		rootNode = parentRuleSet;
+	}
+	
+	markedStrings.push(selectorToSpecificityMarkedString(rootNode));
 	return markedStrings;
 }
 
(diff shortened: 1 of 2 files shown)


aeschli:
You suggest that the fix is to instead of showing the specificity from the current selector, show the one from the top most parent rule set.

That seems only to work fot your test case because the specificity for the top most parent rule set adds the specificity of *all* children rules, which is IMO a bug.

Let's look at this example:
```css
aa {
	background-color: red;
	&.bb { 
		background-color: blue;
	}
	&#cc { 
		background-color: green;
	}
}
```
Hovering over `aa` should return (0,0,1)
Hovering over `bb` should return (0,1,1)
Hovering over `cc` should return (1,0,1)


Here's the test case:
suite('CSS - MarkedStringPrinter selectors', () => {
	test('referencing selector', function() {
		let p = new SCSSParser();
		parseSelectorToMarkedString(p, 'aa { &.bb {} &#cc {} }', 'aa', [
			{ language: 'html', value: '<aa>' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (0, 0, 1)'
		]);
		parseSelectorToMarkedString(p, 'aa { &.bb {} &#cc {} }', 'bb', [
			{ language: 'html', value: '<aa class="bb">' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (0, 1, 1)'
		]);
		parseSelectorToMarkedString(p, 'aa { &.bb {} &#cc {} }', 'cc', [
			{ language: 'html', value: '<aa id="cc">' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (1, 0, 1)'
		]);
		parseSelectorToMarkedString(p, 'aa { .bb {} #cc {} }', 'aa', [
			{ language: 'html', value: '<aa>' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (0, 0, 1)'
		]);
		parseSelectorToMarkedString(p, 'aa { .bb {} #cc {} }', 'bb', [
			{ language: 'html', value: '<aa>\n  …\n    <element class="bb">' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (0, 1, 1)'
		]);
		parseSelectorToMarkedString(p, 'aa { &.bb {} &#cc {} }', 'cc', [
			{ language: 'html', value: '<aa>\n  …\n    <element id="cc">' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (1, 0, 1)'
		]);
		parseSelectorToMarkedString(p, 'o1 { &.c1 { e1 { }}}', 'e1', [
			{ language: 'html', value: '<o1 class="c1">\n  …\n    <e1>' },
			'[Selector Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity): (0, 1, 2)'
		]);
	});
});

aeschli:
So the correct fix is to:
- fix selectorToSpecificityMarkedString to not visit children ruleSets
- fix selectorToSpecificityMarkedString to visit at all parent rules and up the specificities

Alternative it not compute the specificity on the nodes.Node, but on the Element, which is a representation of the hover. To compute the Element, we already wander up the parents and also have some understanding of special selectors (e.g. &).

jonathantisseau:
Hi, this problem still does not seem fixed in version 1.40.0. Can anyone have a look at it?

MindLabor:
> Hi, this problem still does not seem fixed in version 1.40.0. Can anyone have a look at it?

@aeschli Hi this bug is still a thing now in Sep 2020.

aeschli:
@ShPelles Do you still want to work on this?

PR #142, opened 2019-01-08
Allow loading of custom data

octref:
For https://github.com/Microsoft/vscode/issues/64164.

- Refactor `languageFacts`
- Allow loading custom properties/at-directives/pseudo-classes/pseudo-elements
- Strongly type the browser data


src/cssLanguageService.ts
```diff
@@ -22,11 +22,19 @@ import { LESSParser } from './parser/lessParser';
 import { LESSCompletion } from './services/lessCompletion';
 import { getFoldingRanges } from './services/cssFolding';
 import { LanguageSettings, ICompletionParticipant, DocumentContext } from './cssLanguageTypes';
+import { addProperties, addAtDirectives, addPseudoClasses, addPseudoElements, IEntryData } from './services/languageFacts/index';
 
 export type Stylesheet = {};
 export * from './cssLanguageTypes';
 export * from 'vscode-languageserver-types';
 
+export interface LanguageServiceOptions {
```
aeschli: move to 'types'...

src/cssLanguageService.ts
```diff
@@ -72,15 +80,34 @@ function createFacade(parser: Parser, completion: CSSCompletion, hover: CSSHover
...
+		}
+		if (options.customPseudoElements) {
+			addPseudoElements(options.customPseudoElements);
+		}
+	}
+}
 
-export function getCSSLanguageService(): LanguageService {
+export function getCSSLanguageService(options?: LanguageServiceOptions): LanguageService {
+	handleCustomData(options);
```
aeschli: If we choose this API, we must store the properies/directives per LanguageService, not in global arrays. Users can create any number of language services, each with a different set of data.

I would suggest to make a new type `CSSPropertySet` (name to be debated) containing properties/directives/..., that is registered globaly with the `vscode-css-languageservice`. Our built-in properies (from MDN) should also be represented in one (or multiple) CSSPropertySets. 

In a first step, all language services just use that global list of propertySets. In a later step we allow to create a language service with a custom list of Sets. That language service would then only use that set. It would be good to already prepare our code to avoid accessing CSSPropertySet globaly.

src/test/css/customData.test.ts
```diff
@@ -0,0 +1,120 @@
...
+			{
+				name: '::foo',
+				desc: 'Foo pseudo element'
+			}
+		]
+	};
+	return getCSSLanguageService(customData);
+}
+
+function assertCompletion(completions: CompletionList, expected: ItemDescription, document: TextDocument) {
```
aeschli: I'd rather not copy, but just add new tests to completions.tests

src/cssLanguageService.ts
octref: Yeah I was about to check with you. The same problem with HTML now. Do we make the language services stateful? I'll go with stateful then.

src/services/cssCompletion.ts
```diff
@@ -581,7 +581,10 @@ export class CSSCompletion {
 	}
 
 	public getCompletionForTopLevel(result: CompletionList): CompletionList {
-		for (let entry of languageFacts.getAtDirectives()) {
+		const atDirectives = languageFacts.getAtDirectives();
```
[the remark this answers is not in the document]
octref: We need `atDirectives[entryName]` for accessing the each entry. It previously returns an array, so `let...of` works, now it's an object.

PR #143, opened 2019-01-16
fix: handle incomplete rules in box model analysis

yume-chan:
I did't consider rules without values before.

fix Microsoft/vscode#66519


aeschli:
Thanks @yume-chan !

PR #149, opened 2019-03-04
Implement CSS triggerPropertyValueCompletion setting

svipas:
When this PR will be merged, I will create a setting inside `css-language-features` which is in `vscode` repo.

Related to: https://github.com/Microsoft/vscode/issues/68495

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -9,7 +9,7 @@ import { Symbols, Symbol } from '../parser/cssSymbolScope';
 import * as languageFacts from '../languageFacts/facts';
 import * as strings from '../utils/strings';
 import { TextDocument, Position, CompletionList, CompletionItem, CompletionItemKind, Range, TextEdit, InsertTextFormat } from 'vscode-languageserver-types';
-import { ICompletionParticipant, CompletionSettings, LanguageSettings } from '../cssLanguageTypes';
+import { ICompletionParticipant, LanguageSettings } from '../cssLanguageTypes';
 
 import * as nls from 'vscode-nls';
 const localize = nls.loadMessageBundle();
@@ -17,7 +17,7 @@ const SnippetFormat = InsertTextFormat.Snippet;
 
 export class CSSCompletion {
 
-	private settings: CompletionSettings;
+	private settings: LanguageSettings;
 
 	variablePrefix: string;
 	position: Position;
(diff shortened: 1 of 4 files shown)


src/test/css/completion.test.ts
```diff
@@ -209,6 +209,7 @@ suite('CSS - Completion', () => {
 				{ label: 'transition', resultText: 'body { transition:  ' }
 			]
 		});
+		// TODO: how to test it?
```
svipas: @octref Can you help to create tests for this new setting? I tried and it was unsuccessful...

octref: - Make `testCompletionFor` take your completion setting
- In `ItemDescription` add command
- Assert that when you have the setting, retrigger command does not exist on the CompleionItem

src/services/cssCompletion.ts
```diff
@@ -156,7 +162,8 @@ export class CSSCompletion {
 	}
 
 	private getPropertyProposals(declaration: nodes.Declaration, result: CompletionList): CompletionList {
-		let properties = languageFacts.cssDataManager.getProperties();
+		const triggerPropertyValueCompletion = this.settings && this.settings.triggerPropertyValueCompletion;
```
svipas: @octref should I need to check if config exists (undefined) and if it’s undefined return true and if it’s exist take a value? Because right now if we do not pass settings to ‘configure’ or they are undefined, empty, it wouldn’t ‘retrigger’.

octref: You can provide a fallback in the configuration setting (in `css-language-features`) so it's never undefined.
But I agree, `true` should be the default if no setting is found.

svipas:
@octref Could you review again? I created tests and made some small changes.

src/services/cssCompletion.ts
```diff
@@ -32,6 +34,10 @@ export class CSSCompletion {
 		this.variablePrefix = variablePrefix;
 	}
 
+	public configure(settings: LanguageSettings) {
+			this.settings = settings;
```
octref: Spacing

svipas: Good catch, thanks, fixed!

svipas:
@octref Let me know if I can create a PR in VS Code to add `css.completion.triggerPropertyValueCompletion`. But first I need your help to create a description for it 😄

octref:
Something like "By default, VS Code triggers property value completion after selecting a CSS property. Use this setting to disable this behavior." would do.

PR #154, opened 2019-05-04
Fix pseudo-element specificity

parzh:
_See https://github.com/Microsoft/vscode/issues/72608#issuecomment-489332641 for a quick reference_

In CSS3, pseudo-classes (such as `:hover`, `:focus`, or `:invalid`) and pseudo-elements (such as `::before`, `::after` etc.) have some noticeable differences:
- pseudo-classes are more specific than pseudo-elements;
- pseudo-classes are prefixed with `:`, while pseudo-elements start with `::` (though for backwards compatibility this is not enforced);

Currently, they are being distinguished by amount of semicolons, which produces incorrect results when encountering backwards-compatible code:

image

Specificity here is expected to be `(0, 0, 3)`, but got `(0, 1, 2)`. That's because pseudo-element was incorrectly parsed as pseudo-class.

image

Here the number of colons (`::`) corresponds to the entity type (pseudo-element), so the specificity is displayed correctly.

This PR fixes this issue.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,6 @@
   "editor.insertSpaces": false,
   "prettier.useTabs": true,
   "prettier.printWidth": 120,
-  "prettier.semi": true
+  "prettier.semi": true,
+  "typescript.tsdk": "node_modules\\typescript\\lib"
 }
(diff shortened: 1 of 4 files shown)


aeschli:
Thanks @parzh . I modified the RP a bit in order to keep the changes minimal. I hope that's ok for you.

parzh:
@aeschli Thanks for the merge! Yes, that's fine.

PR #159, opened 2019-07-01
Improve scss partial uri building

BPScott:
Currently the regex for checking filenames uses `\w` which only checks for alphanumeric characters and underscore (_). This leads to `@import './foobar'` resolving to `_foobar.scss` while `@import './foo-bar` resolves to `foo-bar`, which is confusing.

This PR makes the prefixing the underscore behaviour more consistent:

- Identifies filenames that contain non-alphanumeric characters,
  'foo-baz' becomes '_foo-baz.scss' and 'foo.bar' becomes '_foo.bar.scss'
- Identifies if a filename is already prefixed with an _, '_foo' becomes
  '_foo.scss' and '_foo.scss' remains '_foo.scss'

Tangentially related: https://github.com/microsoft/vscode/issues/58204 is an issue that points out that the resolution of scss files isn't trivial because `import './foo'` could refer to either `foo.scss` or `_foo.scss` and we need filesystem access to check which of those files exist. So we're still assuming that people should be prefixing their partial's filenames with underscores, but at least this makes that assumption a little more consistent.

### DIFF
--- a/src/services/cssNavigation.ts
+++ b/src/services/cssNavigation.ts
@@ -280,8 +280,8 @@ function uriStringNodeToDocumentLink(document: TextDocument, uriStringNode: node
 }
 
 function toScssPartialUri(uri: string): string {
-	return uri.replace(/\/(\w+)(.scss)?$/gm, (match, fileName) => {
-		return '/_' + fileName + '.scss';
+	return uri.replace(/\.scss$/, '').replace(/[^/]+$/, (basename) => {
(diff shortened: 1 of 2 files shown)


octref:
Thanks, I'll take the test case but I'll rewrite the code with a dynamic resolver.

As I have learned from the Cloudflare news, I'll drop the regex. Especially since this operation is one running on each keystroke.

PR #160, opened 2019-07-03
Rough implementation for scss dynamic resolver

octref:


src/services/cssNavigation.ts
```diff
@@ -104,24 +126,85 @@ export class CSSNavigation {
...
+				originalBasename: uriFragments[uriFragments.length - 1],
+				normalizedUri,
+				normalizedBasename,
+				withBasename(newBaseName: string) {
+					return [...uriFragments.slice(0, -1), newBaseName].join('/');
+				}
+			};
+		}
+		
+		async function fileExists(p: string) {
```
aeschli: Use vcode-URI, uri to file path is not trivial

src/cssLanguageService.ts
```diff
@@ -89,15 +97,39 @@ function handleCustomData(options?: LanguageServiceOptions) {
 
 export function getCSSLanguageService(options?: LanguageServiceOptions): LanguageService {
 	handleCustomData(options);
-	return createFacade(new Parser(), new CSSCompletion(), new CSSHover(), new CSSNavigation(), new CSSCodeActions(), new CSSValidation());
+	return createFacade(
+		new Parser(),
+		new CSSCompletion(),
+		new CSSHover(),
+		new CSSNavigation(),
```
aeschli: I would pass in the fileSystemProvider in the constructor
- no need to add `fsProvider` to the facade

octref:
This is ready for review.

octref:
`findDocumentLinks` now return a promise of `DocumentLink[]` so we might need to publish new major version. Another choice is to instead introduce a `findDynamicDocumentLinks` method so we don't break existing API.

aeschli:
Yes, could do findDocumentLinks2 so it's clear it's the replacement of the old one

PR #166, opened 2019-07-25
Custom data and Markdown support. Fix #132

### LINKED ISSUE #132: Convert to use MarkedString
`MarkedString` is deprecated and causes confusion. See https://github.com/Microsoft/vscode/issues/62332

octref:
- Allow `MarkupContent` in custom data JSON
- Build up hover instead of returning early when a Node has hover content (otherwise all hovers over selector show the specificity instead of the more granular hover)


src/cssLanguageTypes.ts
```diff
@@ -61,9 +61,14 @@ export interface LanguageServiceOptions {
 
 export type EntryStatus = 'standard' | 'experimental' | 'nonstandard' | 'obsolete';
 
+export type MarkupDescription = {
```
aeschli: Why not using the type from vscode-languageserver-types: https://github.com/microsoft/vscode-languageserver-node/blob/db16b7f53b318e8ff44aa95ac8faa5e2486b63aa/types/src/main.ts#L1321

aeschli:
Also I suggest to add a test that shows a hover with both markdown text and specificity.

src/services/cssCompletion.ts
```diff
@@ -914,6 +917,30 @@ export class CSSCompletion {
...
+			return {
+				kind: 'plaintext',
+				value: rawDescription.value
+			};
+		}
+		return rawDescription;
+	}
+
+	private doesSupportMarkdown() {
+		if (!this.clientCapabilities) {
```
aeschli: move this inside the next if (`ifDefined(this.supportsMarkdown)`) ?

src/services/cssHover.ts
```diff
@@ -108,6 +113,52 @@ export class CSSHover {
 			}
 		}
 
+		
+		if (hover) {
+			console.log(this.doesSupportMarkdown());
```
aeschli: remote the log

src/services/cssHover.ts
```diff
@@ -108,6 +113,52 @@ export class CSSHover {
...
 	}
+
+	private convertContents(contents: MarkupContent | MarkedString | MarkedString[]): MarkupContent | MarkedString | MarkedString[] {
+		if (!this.doesSupportMarkdown()) {
+			if (typeof contents === 'string') {
+				return contents;
+			}
+			// MarkupContent
+			else if ('kind' in contents) {
+				return {
```
aeschli: is kind is Markdown it could be returned as  MarkedString

src/services/cssHover.ts
```diff
@@ -108,6 +113,52 @@ export class CSSHover {
...
+					kind: 'plaintext',
+					value: contents.value
+				};
+			}
+			// MarkedString[]
+			else if (isArray(contents)) {
+				contents.map(c => {
```
aeschli: add a return here, just to avoid surprises later

src/services/cssHover.ts
```diff
@@ -108,6 +113,52 @@ export class CSSHover {
...
+				return contents.value;
+			}
+		}
+
+		return contents;
+	}
+
+	private doesSupportMarkdown() {
+		if (!this.clientCapabilities) {
+			this.supportsMarkdown = true;
```
aeschli: default is true?

src/services/cssHover.ts
octref: This is only when `!this.doesSupportMarkdown` is true

src/services/cssHover.ts
octref: I assume if client doesn't specify capabilities we should use whatever that's good for us.

src/services/cssHover.ts
octref: Was outdated change

src/services/cssHover.ts
aeschli: MarkedString was part of the very first LSP, so in theory, it can always be returned

src/services/cssHover.ts
aeschli: In JSON-LS, when no capabilities are passed to the LS constructor, we use the LSP 1.0 capabilities (no MarkupContent, but MarkedString).
So it was done backward compatible (I remembered that wrong)

src/services/cssHover.ts
octref: If client doesn't support MD we should return `{ kind: 'plaintext' }` instead of `MarkedString` which would be interpreted as MD, don't we?

PR #170, opened 2019-08-19
Update mdn-data dependency

bennypowers:
Provides many additional CSS properties e.g. `place-self`


octref:
`place-self` is already available, though:

image

Thanks for sending the PR still.

octref:
@bennypowers I just realized you haven't signed the CLA...Do you mind doing that?

PR #172, opened 2019-08-20
Add more data from MDN and MDN link. Fix #150

### LINKED ISSUE #150: Include MDN links in property IntelliSense
I'm sure I've mentioned this before or discussed it in another issue, but I couldn't find anything specifically about it in the existing issues.

I'd like to be able to hover over, e.g. `background-color` and - besides the description that's already there - have a clickable link to [the MDN documentation](https://developer.mozilla.org/docs/Web/CSS/background-color) for the property.

What it looks like right now:

Screen Shot 2019-04-02 at 12 17 09 PM

The browser-compat-data package includes `mdn_url`s for most (all?) CSS properties, so it should be possible to derive the documentation links from that package, like is already done for browser support.

https://github.com/mdn/browser-compat-data/blob/3268af01c1bf6059f3d7a25c1c212ac4522c291f/css/properties/background-color.json#L6

There's a similar issue for the HTML service: https://github.com/Microsoft/vscode-html-languageservice/issues/12

octref:


src/services/cssCompletion.ts
```diff
@@ -918,15 +918,15 @@ export class CSSCompletion {
...
 
-	private getEntryDescription(entry: languageFacts.IEntry2): string | MarkupContent {
-		const rawDescription = languageFacts.getEntryDescription(entry);
-		if (!this.doesSupportMarkdown() && typeof rawDescription !== 'string' && rawDescription && rawDescription.kind === 'markdown') {
+	private getEntryDescription(entry: languageFacts.IEntry2): string | MarkupContent | undefined {
+		if (this.doesSupportMarkdown()) {
 			return {
-				kind: 'plaintext',
-				value: rawDescription.value
+				kind: 'markdown',
```
aeschli: Let's do it like in the html-ls: 'generateMarkupcontent'. It can generate the MarkupContent or string (based on capabilities)

Issue #173, opened 2019-08-25
Error while computing document links for ...


sryze:
I get this error every time I save a *.scss file:

```
[Error - 1:41:25 AM] Error while computing document links for file:///c%3A/Users/myuser/folder/scss/amp/_header.scss: URI malformed
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at Function.e.parse (c:\Users\myuser\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\css-language-features\server\dist\cssServerMain.js:1:738848)
    at t.<anonymous> (c:\Users\myuser\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\css-language-features\server\dist\cssServerMain.js:1:744433)
    at c:\Users\myuser\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\css-language-features\server\dist\cssServerMain.js:1:743828
    at Object.next (c:\Users\myuser\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\css-language-features\server\dist\cssServerMain.js:1:743933)
    at s (c:\Users\myuser\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\css-language-features\server\dist\cssServerMain.js:1:742702)
```

It causes the bottom (output) panel to pop up every time I hit save and this is quite annoying TBH.

(I'm not sure if this is the correct repo for this bug, should I create it in the vscode repo instead?)

VS Code version:

```
1.37.1
f06011ac164ae4dc8e753a3fe7f9549844d15e35
x64
```

Edit:

1.36 works fine

dbanksdesign:
I am having the same issue. Here is my environment:

Version: 1.38.0
Commit: 3db7e09f3b61f915d03bbfa58e258d6eee843f35
Date: 2019-09-03T21:47:00.464Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Darwin x64 18.7.0

octref:
Do you have a repro case?

alezzzzz:
Same problem here! 
Environment:

Version: 1.38.1 (user setup)
Commit: b37e54c98e1a74ba89e03073e5a3761284e3ffb0
Date: 2019-09-11T13:35:15.005Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
[...]

PR #180, opened 2019-10-05
Wrap URI.parse() call in try ... catch

sryze:
When `URI.parse()` encounters an invalid URL it can throw a `URIError`, but the error is never caught and it pops up in the user's face (issue #173). So this patch simply puts the parse call in a try ... catch statement.

I don't know if the URI library is supposed to throw an error in this case, it may be an issue with vscode-uri.

Steps to reproduce the error:

1. Create a file named test.scss
2. Paste:

```
.test {
    background: url('<%=invalid%>');
}
```

3. Open the file in VS Code and observe an error notification:

error-popup

If you are wondering why I would be using an invalid URL - it's not really invalid, the CSS file produced by the SCSS compiler in my project is included inline into a JSP page, and the `url()` actually contains a JSP echo tag inside (`<%=some expression%>`). This used to work in VS Code 1.36 and earlier, but started to throw an error in newer releases.

### DIFF
--- a/src/services/scssNavigation.ts
+++ b/src/services/scssNavigation.ts
@@ -41,7 +41,15 @@ export class SCSSNavigation extends CSSNavigation {
 					continue;
 				}
 
-				const parsedUri = URI.parse(target);
+				let parsedUri = null;
+				try {
+					parsedUri = URI.parse(target);
+				} catch (e) {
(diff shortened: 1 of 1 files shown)


aeschli:
Looks good, thanks @sryze !

PR #184, opened 2019-10-13
Fix SCSS media query interpolation after 'and'

### LINKED ISSUE #157: SCSS parse error when using Interpolation with mediaQuery
The scss code bellow is grammerly correct and works fine.

``` scss
@mixin res($key, $map: $--breakpoints) {
  @if (map-has-key($map, $key)) {
    @media only screen and #{inspect(map-get($map, $key))} {
      @content;
    }
  } @else {
    @warn "Undefined points: `#{$map}`"
  }
}
```
but VsCode throw a parsing error `( expected scss(css-lparentexpected)`

---

This is on:

- Mac OS v10.14.5
- VsCode v 1.34.0
- Ruby Sass v3.5.7

wongjn:
Fixes #157.

### DIFF
--- a/src/parser/cssParser.ts
+++ b/src/parser/cssParser.ts
@@ -933,6 +933,11 @@ export class Parser {
 			parseExpression = this.acceptIdent('and');
 		}
 		while (parseExpression) {
+			// Allow short-circuting for other language constructs.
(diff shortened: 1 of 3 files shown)


octref:
I rebased on master. Good work. Thanks again!

PR #187, opened 2019-10-24
Add examples for margin and padding css hover

Ragnoroct:
I almost always forget what the values are for the padding property. Having these accessible directly through a hover would be extremely helpful.


octref:
The data is compiled from https://github.com/microsoft/vscode-custom-data/tree/master/web-data, so changes should go into there.

With that said, we need MDN to have metadata on shorthand properties to do this feature.

PR #191, opened 2019-10-28
Fix CSS duplicate semicolon on completion

Pytal:
Closes microsoft/vscode#82726

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -190,15 +190,17 @@ export class CSSCompletion {
 					insertText += ': ';
 					retrigger = true;
 				}
+			} else {
+				range = this.getCompletionRange(null);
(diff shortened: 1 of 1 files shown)


octref:
Sorry this is not the right fix. When you have cases like:

```css
.foo {
  col| ;
}
.foo {
  co|: ;
}
```

These all comes without `semicolonPosition`. It only comes with a complete declaration.

I'll fix this myself. Thanks for trying to help out!

Pytal:
No problem!

I've been digging through trying to understand the codebase.

I look forward to see your fix!

octref:
The fix is in https://github.com/microsoft/vscode-css-languageservice/commit/a4c26e58d69ac3ff554e31cf6abcfd4ce81edcf7, if you are interested.

PR #10, opened 2016-11-15
Add support for parsing @apply rules.

rictic:
Follows the proposed (WIP) spec as it exists today at https://tabatkins.github.io/specs/css-apply-rule/#using

Fixes #8

PR #194, opened 2019-11-20
Add support for parsing @apply rules in SCSS

jorgecasar:
Related with #8 into SCSS files. This remove the error of use `@apply` inside SCSS.


aeschli:
I don't see any mentioning of `@apply` in https://sass-lang.com/documentation/at-rules/. Should it rather be added to css (with a deprecated warning)?

jorgecasar:
It is a deprecated css standard, that is supported in CSS (#10), and doesn't work in SCSS files.

aeschli:
Oh, now I see. Thanks @jorgecasar !

PR #198, opened 2019-11-28
Fix linear gradient description.

peterblazejewicz:
This fixes obvious mistake, aligning with other term usage.

Thanks!


octref:
Thank you!

Issue #203, opened 2019-12-19
[bug] variables before use and forward


MattMcAdams:
Given the following example scss:
```scss
$example: 'random-string';

@use 'test' with (
  $config: $example
);
```
`@use` is flagged with a validation error `at-rule or selector expected` from `scss(css-ruleorselectorexpected)`

You should be allowed to define variables before the `@use` or `@forward` statements for use in module configuration, and the above will compile (given it has real data) using dart sass

From Sass documentation:
> A stylesheet’s `@use` rules must come before any rules other than `@forward`, including style rules. However, you can declare variables before `@use` rules to use when configuring modules.

VS Code version: Code 1.41.0 (9579eda04fdb3a9bba2750f15193e5fafe16b959, 2019-12-11T17:58:38.338Z)
OS version: Darwin x64 19.2.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-5575R CPU @ 2.80GHz (4 x 2800)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off<br>surface_control: disabled_off<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>viz_display_compositor: enabled_on<br>viz_hit_test_surface_layer: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|3, 2, 2|
|Memory (System)|16.00GB (3.72GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (16)</summary>

Extension|Author (truncated)|Version
---|---|---
github-markdown-preview|bie|0.0.2
markdown-checkbox|bie|0.1.3
markdown-emoji|bie|0.0.9
markdown-preview-github-styles|bie|0.1.6
markdown-yaml-preamble|bie|0.0.4
comment-anchors|Exo|1.6.1
php-debug|fel|1.13.0
auto-close-tag|for|0.5.6
auto-rename-tag|for|0.1.1
vscode-scss|mrm|0.8.1
atom-keybindings|ms-|3.0.6
[...]

PR #207, opened 2020-02-11
Fix variables before use and forward

JeremiePat:
This is a proposed fix for Issue #203 

It should work but I'm unsure about throwing the semicolon token part:

In essence it is required if we want to parse all @foward and @use rules because `_parseVariableDeclaration` doesn't throw it away itself (and some other part of the parser expect that behavior). I'm just throwing the semicolon without any further check because it isn't a problem if a file start with a semicolon in SCSS but let me know if we want to be a bit more conservative here.

### DIFF
--- a/src/parser/scssParser.ts
+++ b/src/parser/scssParser.ts
@@ -22,7 +22,11 @@ export class SCSSParser extends cssParser.Parser {
 	}
 
 	public _parseStylesheetStart(): nodes.Node | null {
-		return this._parseForward()
+		if (this.peek(TokenType.SemiColon)) {
+			this.consumeToken();
+		}
+		return this._parseVariableDeclaration()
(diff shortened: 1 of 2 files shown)


JeremiePat:
@aeschli Would it be worth trying to also solve issue #209 it this PR or should I open another one?

aeschli:
Thanks for the PR. I fixed it in a slighly different way through https://github.com/microsoft/vscode-css-languageservice/commit/ef2206162b1cd34618917f40ea65519827800cbb

This no longer enforce in the parser that @use or @forward rule are first.
This check should be done in a linter (likely the scss watcher that users run).

aeschli:
@JeremiePat A (separate) PR for #209 would be highly welcome!

PR #216, opened 2020-05-25
Unknown property: dominant-baseline in svg texts

pguijarr:
## System:
 - OS: Linux 4.15 Ubuntu 18.04.4 LTS (Bionic Beaver)
 - CPU: (2) x64 Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
 - Memory: 881.58 MB / 7.79 GB
 - Container: Yes
 - Shell: 3.1.2 - /usr/bin/fish
## Binaries:
 - Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
 - Yarn: 1.19.1 - ~/.nvm/versions/node/v12.13.0/bin/yarn
 - npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
## npmPackages:
 - styled-components: ^5.1.0 => 5.1.0

## Reproduction

image

## Expected Behavior
Property should be found 
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/dominant-baseline
## Actual Behavior
image


aeschli:
webCustomData is generated from the vscode-web-custom-data NPM package (https://github.com/microsoft/vscode-custom-data)...

dominant-baseline is a svg attribute and we don't have them in our cutom data yet.
IMO the best would eb that we generate we should generate a separate css-data fiel just for the SVG attributes.
That would happen in vscode-web-custom-data: Something like `vscode-web-custom-data/data/svg.css-data.json`

PR #220, opened 2020-09-07
chore(deps): vscode-nls@5.0.0

AviVahl:
contains browser compatibility fixes. see: https://github.com/microsoft/vscode-nls/pull/28


aeschli:
Thanks @AniVahl!

PR #222, opened 2020-10-05
[scss] correctly identify an element with id (div#id)

ShPelles:
fix microsoft/vscode#107269


aeschli:
Thanks a lot, @ShPelles !

PR #228, opened 2020-12-08
Use node type to check for id selectors

aNickzz:
Following on from the discussion in https://github.com/microsoft/vscode/issues/75905#issuecomment-733606749.

This change replaces the check for a `#` character with a node type check when linting ID selectors. This is to avoid the "avoid id selectors" rule from highlighting interpolations in scss.

This also fixes a bug that the ID lint would only warn when the ID was the first selector in the rule, eg. it would detect `#foo.bar` but not `.bar#foo`.

I've added a couple more tests to cover some different id selectors and interpolations.

While this works, would it be better to add a condition on [this switch](https://github.com/microsoft/vscode-css-languageservice/blob/982dc8849a81e0f77c12b57c34e2da5263b93f2e/src/services/lint.ts#L152) to find and handle `IdentifierSelectors` rather than the loop through the `SimpleSelector` children that I have in this PR?

### DIFF
--- a/src/services/lint.ts
+++ b/src/services/lint.ts
@@ -4,15 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 'use strict';
 
-import * as languageFacts from '../languageFacts/facts';
-import { Rules, LintConfigurationSettings, Rule, Settings } from './lintRules';
-import * as nodes from '../parser/cssNodes';
-import calculateBoxModel, { Element } from './lintUtil';
-import { union } from '../utils/arrays';
-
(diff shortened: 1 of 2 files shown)


aeschli:
Looks good! I like the suggestion to add a function `visitIdentifierSelectors`. It's better not to assume that an IdentifierSelectors is always the direct child of a SimpleSelector

aNickzz:
Much simpler

aeschli:
Thanks @aNickzz !

PR #254, opened 2021-10-24
Respect `kind` property

avigoldman:
The current code escapes the markdown regardless of if the `kind` property is set to `markdown`. 

This PR updates it to skip escaping the string if it is supposed to be markdown.


aeschli:
Thanks @avigoldman !

PR #255, opened 2021-10-27
Feat: support new resolving import at-rules

fyangstudio:
https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules

> Using `~` is deprecated and can be removed from your code (we recommend it), but we still support it for historical reasons. Why can you remove it? The loader will first try to resolve @import as a relative path. If it cannot be resolved, then the loader will try to resolve @import inside `node_modules`.

Support VS Code DefinitionProvider find right package like:

```scss
@import "bootstrap";
@import "~bootstrap";
```

### DIFF
--- a/src/services/cssNavigation.ts
+++ b/src/services/cssNavigation.ts
@@ -303,7 +303,7 @@ export class CSSNavigation {
 				return joinPath(modulePath, pathWithinModule);
 			}
 		}
-		return documentContext.resolveReference(ref, documentUri);
+		return undefined;
 	}
 
 	protected async resolveRelativeReference(ref: string, documentUri: string, documentContext: DocumentContext, isRawLink?: boolean): Promise<string | undefined> {
@@ -322,7 +322,7 @@ export class CSSNavigation {
 		if (relativeReference) {
 			return relativeReference;
 		} else {
-			return await this.resolveModuleReference(ref, documentUri, documentContext)
+			return await this.resolveModuleReference(ref, documentUri, documentContext);
 		}
 	}
 
--- a/src/services/cssNavigation.ts
+++ b/src/services/cssNavigation.ts
@@ -297,7 +297,7 @@ export class CSSNavigation {
 		// and [sass-loader's](https://github.com/webpack-contrib/sass-loader#imports)
 		// convention, if an import path starts with ~ or a package name then use node module resolution
 		// *unless* it starts with "~/" as this refers to the user's home directory.
-		if (/[a-zA-Z0-9~]/.test(ref[0]) && ref[1] !== '/' && this.fileSystemProvider) {
+		if (/[a-zA-Z0-9_~]/.test(ref[0]) && ref[1] !== '/' && this.fileSystemProvider) {
 			ref = ref.substring(1);
 			if (startsWith(documentUri, 'file://')) {
 				const moduleName = getModuleNameFromPath(ref);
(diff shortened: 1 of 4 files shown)


fyangstudio:
@aeschli

aeschli:
> The loader will first try to resolve @import as a relative path. If it cannot be resolved, then the loader will try to resolve @import inside node_modules

The looking up of an import in node_modules, Is this something webpack does, or SCSS?

fyangstudio:
> > The loader will first try to resolve @import as a relative path. If it cannot be resolved, then the loader will try to resolve @import inside node_modules
> 
> The looking up of an import in node_modules, Is this something webpack does, or SCSS?

Webpack and Sass are both mentioned that imports will always be resolved relative to the current file first, see: 

Webpack: https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules
Sass: https://sass-lang.com/documentation/at-rules/import#load-paths

This PR support node module resolution without `~`. 

PS: using `~` is work too

fyangstudio:
> > The loader will first try to resolve @import as a relative path. If it cannot be resolved, then the loader will try to resolve @import inside node_modules
> 
> The looking up of an import in node_modules, Is this something webpack does, or SCSS?

@aeschli   Webpack and Sass are both mentioned that imports will always be resolved relative to the current file first, see: 

Webpack: https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules
Sass: https://sass-lang.com/documentation/at-rules/import#load-paths

This PR support node module resolution without `~`. 

PS: using `~` is work too

aeschli:
Thanks @fyangstudio.
I wonder if we should only enable this for SCSS. Right now, that change would enable the extended resolving also for LESS and CSS.
For CSS there's the [css-loader](https://github.com/webpack-contrib/css-loader#url). If I get this right, it still requires '~'.

Also, it would be nice to have tests as well.

src/services/cssNavigation.ts
```diff
@@ -292,26 +292,38 @@ export class CSSNavigation {
...
-					return joinPath(modulePath, pathWithinModule);
-				}
-			}
-			return documentContext.resolveReference(ref, documentUri);
+			return await this.resolveModuleReference(ref, documentUri, documentContext) || relativeReference;
+		}
+
+		// Using ~ is deprecated. The loader will first try to resolve @import as a relative path. If it cannot be resolved, 
+		// then the loader will try to resolve @import inside node_modules.
+		if (relativeReference) {
```
aeschli, line 322: This is always true. documentContext.resolveReference just computes a URI, but does not make an existence test.

aeschli (changes requested):

fyangstudio:
> Thanks @fyangstudio. I wonder if we should only enable this for SCSS. Right now, that change would enable the extended resolving also for LESS and CSS. For CSS there's the [css-loader](https://github.com/webpack-contrib/css-loader#url). If I get this right, it still requires '~'.
> 
> Also, it would be nice to have tests as well.

@aeschli   '~' is deprecated in Less too,  https://webpack.js.org/loaders/less-loader/ .

I think css-loader will support the new resolving import at-rules soon.

fyangstudio:
> Thanks @fyangstudio. I wonder if we should only enable this for SCSS. Right now, that change would enable the extended resolving also for LESS and CSS. For CSS there's the [css-loader](https://github.com/webpack-contrib/css-loader#url). If I get this right, it still requires '~'.
> 
> Also, it would be nice to have tests as well.

'~' is deprecated in Less too,  https://webpack.js.org/loaders/less-loader/#webpack-resolver 

I change it only enable new rule for Sass and Less,  @aeschli  please check.

By the way I think css-loader will support the new resolving import at-rules soon.

fyangstudio:
@aeschli if this PR have any questions, please let me know.

aeschli:
I polished the check whether to resolve module references (https://github.com/microsoft/vscode-css-languageservice/pull/255/commits/6e64249552bd96049f92e174d6ff7d7c35a7354b)

All looks good now, thanks @fyangstudio !

aeschli (approved):

PR #257, opened 2021-12-23
css color : add support for hue rad|grad|turn

romainmenke:
[spec](https://www.w3.org/TR/css-values-4/#angle-value)

> 7.1. Angle Units: the <angle> type and deg, grad, rad, turn units
> 
> Angle values are <dimension>s denoted by <angle>. The angle unit identifiers are:
> 
> deg
> Degrees. There are 360 degrees in a full circle.
>
> grad
> Gradians, also known as "gons" or "grades". There are 400 gradians in a full circle.
>
> rad
> Radians. There are 2π radians in a full circle.
>
> turn
> Turns. There is 1 turn in a full circle.

-------

Aiming to get the color indicator here :

```
color: hsl(0.5turn 100% 50%);
color: hsl(200grad 100% 50%);
color: hsl(3.14159rad 100% 50%);
```

Screenshot 2021-12-23 at 10 45 19

### DIFF
--- a/src/languageFacts/colors.ts
+++ b/src/languageFacts/colors.ts
@@ -196,11 +196,11 @@ function getAngle(node: nodes.Node) {
 			case 'deg':
 				return parseFloat(val) % 360;
 			case 'rad':
-				return parseFloat(val) * 180 / Math.PI;
+				return (parseFloat(val) * 180 / Math.PI) % 360;
 			case 'grad':
-				return parseFloat(val) * 0.9;
+				return (parseFloat(val) * 0.9) % 360;
 			case 'turn':
-				return parseFloat(val) * 360;
+				return (parseFloat(val) * 360) % 360;
 			default:
 				if ('undefined' === typeof m[2]) {
 					return parseFloat(val) % 360;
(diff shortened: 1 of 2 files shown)


romainmenke:
@aeschli could you take a look when you have time?

Happy to make any needed changes :)

src/languageFacts/colors.ts
```diff
@@ -190,10 +190,24 @@ function getNumericValue(node: nodes.Node, factor: number) {
...
 	const val = node.getText();
-	const m = val.match(/^([-+]?[0-9]*\.?[0-9]+)(deg)?$/);
+	const m = val.match(/^([-+]?[0-9]*\.?[0-9]+)(deg|rad|grad|turn)?$/);
 	if (m) {
-		return parseFloat(val) % 360;
+		switch (m[2]) {
+			case 'deg':
+				return parseFloat(val) % 360;
+			case 'rad':
+				return parseFloat(val) * 180 / Math.PI;
```
aeschli, line 199: For 'rad', 'grad' and 'turn', can they represent multiple turns? Should we add a modulo 360 as well?

romainmenke: 🤔 Will check.
Good catch!

romainmenke: This has been updated.

aeschli:
Cool thanks for the PR and the quick update!

PR #262, opened 2022-02-08
add support for hwb() color function

romainmenke:
docs : https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hwb()
spec : https://www.w3.org/TR/css-color-4/#the-hwb-notation
based on : https://github.com/w3c/csswg-drafts/blob/main/css-color-4/hwbToRgb.js

- `yarn test` passes
- verified with `css-language-features` and manually ran the extension

I did not find how to localise `Creates a Color from hue, white and black.`.
Is this done elsewhere?

Screenshot 2022-02-08 at 23 38 36

------

Also opened a PR to get better highlighting for `hwb()` : https://github.com/microsoft/vscode/pull/142536


aeschli:
Very cool, thanks a lot, @romainmenke !

I added some more tests.

I'm not sure what you mean by
`I did not find how to localise Creates a Color from hue, white and black..`

You mean the actual translation? It's done in https://github.com/microsoft/vscode-loc it's a gap that we don't translate strings contained in the language servers.

romainmenke:
Yes that is what I was looking for :)
I mainly didn't want to create new gaps if everything else was translated.

Thank you for the extra tests and the review!

Issue #248, opened 2021-09-13
Wrong selector specificity


Ayc0:
When looking at the specify VSCode gives us for the selector `a:where(a)`, we can see that it shows `(0, 1, 2)` (because 2 `a` (0, 0, 1) and 1 `:where` (0, 1, 0)).

![VSCode showing a specificity of `(0, 1, 2)` for the selector `a:where(a)`](https://user-images.githubusercontent.com/22725671/133162556-8111fca2-075d-412f-b8ed-48dddafefc51.png)

But actually, `:where` doesn't add any specificity, so it should return `(0, 0, 1)`. [`Parsel`](https://projects.verou.me/parsel/?selector=a%3Awhere%28a%29) confirms this:

![Parsel showing a specificity of `(0, 0, 1)` for the selector `a:where(a)`](https://user-images.githubusercontent.com/22725671/133162801-0854d9df-70ba-4552-8bf4-12e2a838458a.png)

PR #263, opened 2022-02-08
fix specificity for :has, :not, :is and :where

romainmenke:
fixes : https://github.com/microsoft/vscode-css-languageservice/issues/248

- `:has`, `:is` and `:not` now get the specificity of the most specific list item.
- `:where` always adds zero specificity

Screenshot 2022-02-09 at 09 19 07

Screenshot 2022-02-09 at 00 18 27


aeschli:
@romainmenke Cool, thanks a lot!

romainmenke:
Needed it in postcss-preset-env, also suggesting it here was a small effort :)

Thank you for reviewing and merging so quickly!

PR #268, opened 2022-03-31
fix selector specificity display for `:lang` and similar pseudo selectors

romainmenke:
Screenshot 2022-03-31 at 17 52 40

https://drafts.csswg.org/selectors/#specificity-rules

> A few pseudo-classes provide “evaluation contexts” for other selectors, and so have their specificity defined specially:

With this change only those selectors with an `evaluation context` will have their child nodes counted.

### DIFF
--- a/build/copy-jsbeautify.js
+++ b/build/copy-jsbeautify.js
@@ -16,14 +16,6 @@ function copy(from, to) {
 	}
 }
 
-if (!fs.existsSync(path.join(__dirname, '..', 'lib', 'umd'))) {
-	fs.mkdirSync(path.join(__dirname, '..', 'lib', 'umd'));
-}
-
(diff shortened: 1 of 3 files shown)


build/copy-jsbeautify.js
```diff
@@ -16,6 +16,14 @@ function copy(from, to) {
 	}
 }
 
+if (!fs.existsSync(path.join(__dirname, '..', 'lib', 'umd'))) {
```
romainmenke, line 19: This isn't needed to fix tests, but included to mirror the fix for `esm`

aeschli: This should be fixed by e6deb0b91579ee9e6d34893f5317559c22ba5710

romainmenke: Awesome! I'll update my PR!

aeschli:
Thanks @romainmenke !

PR #270, opened 2022-04-16
sass-forward rule keyword "with"

mariamhmahmoud:
This PR works to resolve https://github.com/microsoft/vscode/issues/145108. The implementations looks to see if "with" follows keyword "as" and resolves the "semicolon expected" error present in the issue.


aeschli:
I came up with #294 (very similar to your fix)

PR #276, opened 2022-06-10
feat: support for css cascade layer at-rule

praveenpuglia:
I obviously don't understand much of parsing logic here but I did the bits I could do.
This current PR at least takes the squiggly off of editor and doesn't report it as an error. 

Would love to know how to write the parser for this.
- [x] Add`@layer` to the web data. 
- [ ] parser for `@layer`
- [ ] `layer()` function support in `@import`;

### DIFF
--- a/src/data/webCustomData.ts
+++ b/src/data/webCustomData.ts
@@ -20019,6 +20019,17 @@ export const cssData : CSSDataV1 = {
 			"description": "Defines set of animation key frames."
 		},
 		{
+			name: "@layer",
+			references: [
+				{
+					name: "MDN Reference",
+					url: "https://developer.mozilla.org/docs/Web/CSS/@layer",
+				},
+			],
+			description:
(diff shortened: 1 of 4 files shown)


src/data/webCustomData.ts
```diff
@@ -1,11 +1,11 @@
-/*---------------------------------------------------------------------------------------------
- *  Copyright (c) Microsoft Corporation. All rights reserved.
- *  Licensed under the MIT License. See License.txt in the project root for license information.
- *--------------------------------------------------------------------------------------------*/
-// file generated from @vscode/web-custom-data NPM package
```
praveenpuglia, line 5: Looked at it pretty late! Will make changes there.

aeschli:
I implemented the request through #292

aeschli:
@praveenpuglia Thanks for the help!

PR #279, opened 2022-06-30
selectors may start with combinators

### LINKED ISSUE #250: Add support for `<relative‑selector‑list>` in the `:has(…)` pseudo‑class
The `:has(…)` pseudo‑class, which is being prototyped in **Blink**, is specified as taking a [`<relative‑selector‑list>`](https://drafts.csswg.org/selectors/#typedef-relative-selector-list).

## See also
- **Chromium** [**Issue 669058:** CSS selectors Level 4: support :has()](https://crbug.com/669058 "Issue 669058: CSS selectors Level 4: support :has()")
- <https://github.com/Igalia/explainers/tree/main/css/has>

---

More broadly, the function syntax for [`<pseudo‑class‑selector>`](https://drafts.csswg.org/selectors/#typedef-pseudo-class-selector) is specified as taking [`<any‑value>`](https://drafts.csswg.org/css-syntax/#typedef-any-value), meaning that the current parser is too restrictive, and also too lax, as `:is(1 - 2)` should result in a warning, since `1 - 2` is not a valid selector, but because `:is(…)` takes a [`<forgiving‑selector‑list>`](https://drafts.csswg.org/selectors/#forgiving-selector), it’s equivalent to `:is()` at runtime, which is a valid selector that matches nothing.

The correct implementation would be to parse pseudo‑class arguments as `<any‑value>`, rather than as <code>&#91; [&lt;complex‑selector‑list&gt;](https://drafts.csswg.org/selectors/#typedef-complex-selector-list) [&#124;](https://drafts.csswg.org/css-values/#comb-one) [&lt;calc‑sum&gt;](https://drafts.csswg.org/css-values/#calc-syntax) &#93;[?](https://drafts.csswg.org/css-values/#mult-opt)</code> as it currently does, and then perform stricter validation of the arguments based on the pseudo‑class/pseudo‑element itself.

taroken6:
fixes https://github.com/microsoft/vscode/issues/147353  
fixes #250

A minor change that removes syntax errors when selectors start with combinators (ex: `>`, `~`, and `+`).

### DIFF
--- a/src/parser/cssParser.ts
+++ b/src/parser/cssParser.ts
@@ -1413,10 +1413,10 @@ export class Parser {
 			if (!this.hasWhitespace() && this.accept(TokenType.ParenthesisL)) {
 				const tryAsSelector = () => {
 					const selectors = this.create(nodes.Node);
-					if (!selectors.addChild(this._parseSelector(false))) {
+					if (!selectors.addChild(this._parseSelector(true))) {
 						return null;
 					}
-					while (this.accept(TokenType.Comma) && selectors.addChild(this._parseSelector(false))) {
+					while (this.accept(TokenType.Comma) && selectors.addChild(this._parseSelector(true))) {
 						// loop
 					}
 					if (this.peek(TokenType.ParenthesisR)) {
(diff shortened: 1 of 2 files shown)


dbaeumer (approved):

aeschli:
@taroken6 Thanks a lot, I added some tests. 
Can you sign the Contributor License Agreement?

taroken6:
> @taroken6 Thanks a lot, I added some tests. Can you sign the Contributor License Agreement?

Think I signed it. Didn't see word from CLA Bot.

PR #285, opened 2022-08-14
fix: handle scoped packages with _index on package root 

### LINKED ISSUE #284: Link error for scoped packages that export from the root level
Hi, and thanks for all the work on vscode and this package – much appreciated.

I maintain an extension that uses this languageservice as a foundation. A user reported a bug to me that turned out to be an issue in the way `findDocumentLinks2` handles scoped NPM packages.

In this case the user maintains an SCSS library that is published in individual packages that, simplified, look like this:

```
@scope/package-name/
    _index.scss
    other.scss
    package.json
```

They use the index to have `@use` statements that look like `@use “@scope/package-name"`.

The function [getModuleNameFromPath in cssNavigation](https://github.com/microsoft/vscode-css-languageservice/blob/045b4764e36648a29cb3970e6733b35a0085600d/src/services/cssNavigation.ts#L423-L425) doesn’t handle these paths. It always expects two forward slashes whenever there’s a scoped module. I applied a local patch to handle this case, and it seems to fix the most important problems my user was having.

```diff
 function getModuleNameFromPath(path) {
-    // If a scoped module (starts with @) then get up until second instance of '/', otherwise get until first instance of '/'
+    // If a scoped module (starts with @) then get up until second instance of '/', or to the end of the string for root-level imports.
     if (path[0] === '@') {
-        return path.substring(0, path.indexOf('/', path.indexOf('/') + 1));
+        const secondSlash = path.indexOf('/', path.indexOf('/') + 1);
+        if (secondSlash === -1) {
+            return path;
+        }
+        return path.substring(0, path.indexOf('/', path.indexOf('/') + 1));
     }
-    return path.substring(0, path.indexOf('/'));
+    // Otherwise get until first instance of '/'
+    return path.substring(0, path.indexOf('/'));
 }
```

Unfortunately, Ctrl/Cmd + Click on the path in the use-statement still doesn’t work with these changes. I've tried to debug the problem, but I don't know how to hook into the `openerService` in `vscode` to solve this.

wkillerud:
Fixes #284 

Like I mentioned in the issue, there's still a bug when I cmd/ctrl + click the link `@use "@foo/bar";` where VS Code tries to open the file of that name. I wasn't able to figure out where and how to fix that.


aeschli:
Thanks @wkillerud. I added anothe commy to fix open. The reason was a unnecessary encoding of the `@`.

I updated the test in the following way
- `@use '@foo/bar/baz` resolves to `node_modules/@foo/bar/_baz.scss`
- `@use '@foo/bar` resolves to `node_modules/@foo/bar/_index.scss`

Please speak up, if I misunderstand how `@use` works

dbaeumer (approved):

wkillerud:
Nice, thank you!

> `@use '@foo/bar/baz` resolves to `node_modules/@foo/bar/_baz.scss`

If `_baz.scss` exists, yes. Otherwise it might resolve to:

- `node_modules/@foo/bar/baz.scss` (non-[partial](https://sass-lang.com/documentation/at-rules/use#partials))
- `node_modules/@foo/bar/baz/_index.scss` (where `baz` is another directory with an [index file](https://sass-lang.com/documentation/at-rules/use#index-files)))

> `@use '@foo/bar resolves` to `node_modules/@foo/bar/_index.scss`

Yes

aeschli:
> If `_baz.scss` exists, yes. Otherwise it might resolve to:
> 
> * `node_modules/@foo/bar/baz.scss` (non-[partial](https://sass-lang.com/documentation/at-rules/use#partials))
> * `node_modules/@foo/bar/baz/_index.scss` (where `baz` is another directory with an [index file](https://sass-lang.com/documentation/at-rules/use#index-files)))

Yes, that's how it's implemented.

PR #297, opened 2022-09-09
relative path locations wrong when using @use

aeschli:
Fixes https://github.com/microsoft/vscode/issues/142836
Fixes https://github.com/microsoft/vscode/issues/153237


dbaeumer (approved):

bencresty:
Thanks a lot for this guys! Great to see you found a way to fix this!

PR #299, opened 2022-09-17
🔨 Include both `<sth>` and `<element name="sth">` in hover

### LINKED ISSUE #160987: Wrongly shown selector in SCSS
image
image

In the first picture it should also show `<element name="jmeno">` like in the second one.

babakks:
Fixes microsoft/vscode#160987


babakks:
@aeschli I used `|` (pipe) for as a joiner for possible element signatures, like:

```
<something>|<element name="something">
```

Of course, this change of behavior of the `MarkedStringPrinter` class resulted in other tests failing. I wasn't sure that you're okay with the changes that I made in this PR. So, I left other test cases as they were, until hearing your opinion on this.

aeschli:
Thanks @babakks. I'm sorry, but I don't think that's the right fix.
`[name='something']` should give the hover `<element name="something">` 
`<something>` is wrong and is the bug.

Note that `<element name="something">` is not something you can write in HTML. The `element` is just used as a placeholder for any element name.

The bug is that the `Element` class uses the `name property to store the element name, but name is a valid attribute, so that conflicts. Instead of name, the element name should be stored in a separate property, or something like `<>` should be used which is not a valid attribute name.

aeschli:

babakks:
No worries, @aeschli! :pray: It's been a long ago and I don't remember a thing in this one.

PR #301, opened 2022-10-20
feat: implement `itemDefaults.editRange`

### LINKED ISSUE #265: Provide `textEdit` in completion item
Currently vscode-css-languageservice do not provide textEdit in completion item, scss document completion text replace range depends on wordPattern from vscode scss language configuration.

In vanilla scss document, it's working good, but if scss is embedded language in a document, vscode don't know the completion item is for scss language, so vscode don't apply replace range by scss wordPattern. In this case vscode can't show expected scss completions.

螢幕截圖 2022-02-13 06 35 17
 
To fix this problem, I was calculate `textEdit` for css completion item before send it to language client.

https://github.com/johnsoncodehk/volar/blob/573dad2efb6faa5fcb68e147498f1a2e58f65b3e/packages/vscode-vue-languageservice/src/services/completion.ts#L649-L666

But this way is unreliable, I've had some particular edge cases before. (But I forgot what actually happen)
I hope we can have `textEdit` from vscode-css-languageservice, so as downstream we don't need to care about css completion replace rules.

johnsoncodehk:
Fix #265

Implement `itemDefaults.editRange` to support at-rules completion in embedded language.


aeschli (approved):

aeschli:
Thanks @johnsoncodehk !

dbaeumer (approved):

Issue #249, opened 2021-09-17
Show multiple elements suggestions


m0ksem:
```scss
input {
  &:hover, &:focus {
    &::placeholder {
      color: red;
    } 
  }
}
```

![image](https://user-images.githubusercontent.com/23530004/133849494-264917e4-bc1f-4683-acd9-65bd464f3b6b.png)

For example show it as:

`<input :hover ::placeholder> or <input :focus ::placeholder>`

or

```html
<input :hover ::placeholder>
<input :focus ::placeholder>
```

Codepen: https://codepen.io/m0ksem/pen/qBjxxzo

GauravB159:
@aeschli Could I look into this issue? I'm a bit familiar with the CSS Language service after working on a couple PRs. 

Also, if this is implemented, would the selector specificity necessarily be the same for multiple element suggestions? One might have a different specificity than the other, how would we handle that?

aeschli:
@GauravB159 Of course, that would be reat. I think each variant would need its own specificity.

Code is in https://github.com/microsoft/vscode-css-languageservice/blob/main/src/services/cssHover.ts

The `hover. Content` is an array and can contain any number of parts. Maybe that helps.

GauravB159:
Awesome, I'll look into it, thank you! @aeschli For the specificity, would displaying them in the same order as the list of elements be enough context to understand which specificity maps to which element?

aeschli:
@GauravB159 For you to experiment. Maybe it can look like that:
```
<input :hover ::placeholder>
Selector specificity: (0.0.1)
<input :focus ::placeholder>
Selector specificity: (0.0.1)
```
If they are the same, maybe one line (at the bottom) is enough.

GauravB159:
@aeschli Gotcha, I'll try it out, thank you.

GauravB159:
@aeschli I'm testing out the changes for this and I noticed MarkedString is supposed to be deprecated. Does that need to be replaced with something else? 

![image](https://user-images.githubusercontent.com/22005033/200531158-6f63b80d-b370-4a68-9430-3d1e40eadf55.png)

m0ksem:
Hi @GauravB159, thanks for dealing with this issue. This looks cool already.

ID should have 1.0.0 specificity according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#selector_weight_categories).

:hover should add 0.1.0 to specificity. It looks like parent selector specificity doesn't count here.
[...]

PR #308, opened 2022-11-08
Added multiple element preview on hover of scss selector

GauravB159:
Partial fix for #249

Still need to figure out the logic for specificity. There seems to be some kind of bug in the production build itself. 

Still need to add tests for the new feature.


GauravB159:
image

Such cases have not been handled yet, only the case where an element's immediate parents are multiple has been handled. I'm not sure if we want to handle this case, since it might just get too complicated. If needed, I can possibly extend it to cover this as well using a recursive method.

aeschli:
Yes, I agree that this can get very complicated. The hover size should not get too long. Maybe write (1 more) behind `<element class="div-two">` ?

GauravB159:
Hmm, I'm not sure how to limit the depth. Possibly depends on what the most frequent use case for this would be. I don't think people would want to preview a really deep node list anyways. 

I'm not sure what you mean by write one more behind `<element class="div-two">`?

Basically, the way it currently works is, if the second to last nested selector has multiple values, it picks up those, but other than that it picks the first selector in each ancestor. 

Basically a depth of 1, I guess.

aeschli:

Issue #309, opened 2022-11-09
[SCSS] Specificity values for nested rules are incorrect


GauravB159:
Discovered this issue while working on #249

![200559512-a3643c9c-e465-4dc2-aefc-9cb242d641e8](https://user-images.githubusercontent.com/22005033/200765572-c9cdd903-597f-4d07-89d4-48529a661869.png)

`input { #id { p { color: red; } } }` - Specificity (0, 0, 1)

`input #id p { color: red; }` - Specificity (1, 0, 2)

Even though they are equivalent

PR #311, opened 2022-11-09
Nested specificity fix

GauravB159:
Fix for issue #309 

All tests were passing except one, which it seems was using the old specificity value in the assertion. I have changed that to the correct value.


aeschli:
Can you add test cases that cover the new code?

GauravB159:
Yes, I'll get on that today

aeschli:

Issue #312, opened 2022-11-10
support media queries level 4 range for SCSS


yukulele:
Same as #213 but for SCSS

```scss
main {
  background: blue;
  @media (width <= 1024) {
    background: green;
  }
}
```
error:
```
) expected scss(css-rparentexpected)
```

GauravB159:
The CSS Scanner picks up `<` and '=` as two separate tokens whereas the SCSS Scanner picks up `<=` as a single token, which might be causing this issue. I tried adding a separate condition for the SCSS Scanner and the error goes away, but the number after does not get highlighted anymore.  I'll link a draft pull request just in case I don't figure it out and someone can pick up from there.

PR #313, opened 2022-11-11
Added condition for SCSS Media Query range

GauravB159:
Partial fix for #312 

The error is gone but the number after the operator does not get highlighted like in CSS

~~Edit: Nevermind, still has an error in the code itself, build will fail currently~~


aeschli:
Fixed by https://github.com/microsoft/vscode-css-languageservice/issues/312

Issue #314, opened 2022-11-13
Color completion only displays color squares for hex variables


GauravB159:
Based on issue microsoft/vscode#166197

I have been checking this out and have found a couple issues. Part of this can be fixed in [microsoft/vscode-css-languageservice](https://github.com/microsoft/vscode-css-languageservice)

![image](https://user-images.githubusercontent.com/22005033/201510834-060809bc-9bbe-419c-a245-cbe1ad6c7782.png)

[cssCompletion.ts](https://github.com/microsoft/vscode-css-languageservice/blob/main/src/services/cssCompletion.ts)

This function only tests for built-in colors and hex values. Adding another part to it that checks for color functions fixes the two color functions.

Making this change fixes $rgb and $rgba.

allan2:
@GauravB159 do you know if your PR also helps with vscode [#166201](https://github.com/microsoft/vscode/issues/166201)?

GauravB159:
@allan2 That's actually a separate issue. I've looked into that as well, and I'm mostly done with it. But there are a few tests failing. I might push a draft PR in a bit anyways because I'm not sure if those tests are based on correct behavior or not.

PR #315, opened 2022-11-13
Fix check for non-hex values

GauravB159:
Partial fix for microsoft/vscode#166197 (#314) in conjunction with microsoft/vscode#166220

Added the condition to check for CSS color functions and moved the function to a more appropriate file for future changes. In case any functions are added, it will be easy to modify this function as well.


aeschli:
Implemented by https://github.com/microsoft/vscode-css-languageservice/pull/346

PR #318, opened 2022-11-22
release pipeline: add quality

aeschli:


lszomoru (approved):

lszomoru:
@aeschli, that is correct. That environment variable is set in the template itself so you can just use it.

Issue #344, opened 2023-04-29
IE star hacks are incompatible with CSS Nesting


romainmenke:
https://github.com/microsoft/vscode-css-languageservice/blob/6db83e5329520a96054ae880a43c04c6ce51da05/src/test/css/lint.test.ts#L232-L235

`*` should start rule parsing for css nesting.

Maybe it is safe to remove the custom handling for IE star hacks, given the age of the browsers targeted by these hacks.

romainmenke:
These tests are now "skipped" : https://github.com/microsoft/vscode-css-languageservice/blob/45f080571333ed82b4b9d4ab5ecfbf59b9de666f/src/test/css/lint.test.ts#L232-L236

PR #345, opened 2023-04-29
css nesting : increase support

romainmenke:
- handle nested at rules
- improve autocomplete and diagnostics when starting a nested rule
- improve parsing behaviour when a rule doesn't start with an ident

### DIFF
--- a/src/parser/cssParser.ts
+++ b/src/parser/cssParser.ts
@@ -320,7 +320,7 @@ export class Parser {
 			|| this._parseFontFace()
 			|| this._parseKeyframe()
 			|| this._parseSupports(isNested)
-			|| this._parseLayer()
+			|| this._parseLayer(isNested)
 			|| this._parsePropertyAtRule()
 			|| this._parseViewPort()
 			|| this._parseNamespace()
@@ -361,7 +361,10 @@ export class Parser {
 	}
 
 	protected _parseRuleSetDeclarationAtStatement(): nodes.Node | null {
-		return this._parseUnknownAtRule();
+		return this._parseMedia(true)
+			|| this._parseSupports(true)
+			|| this._parseLayer(true)
+			|| this._parseUnknownAtRule();
 	}
 
 	public _parseRuleSetDeclaration(): nodes.Node | null {
@@ -369,8 +372,12 @@ export class Parser {
 		if (this.peek(TokenType.AtKeyword)) {
 			return this._parseRuleSetDeclarationAtStatement();
 		}
-		return (!this.peek(TokenType.Ident) && this._tryParseRuleset(true))
-			|| this._parseDeclaration();
+
+		if (!this.peek(TokenType.Ident)) {
+			return this._parseRuleset(true);
+		}
+
+		return this._tryParseRuleset(true) || this._parseDeclaration();
 	}
 
 	public _needsSemicolonAfter(node: nodes.Node): boolean {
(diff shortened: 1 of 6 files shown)


src/test/css/lint.test.ts
```diff
@@ -230,8 +230,9 @@ suite('CSS - Lint', () => {
 	});
 
 	test('IE hacks', function () {
-		assertRuleSet('selector { display: inline-block; *display: inline; }', Rules.IEStarHack);
-		assertRuleSet('selector { background: #00f; /* all browsers including Mac IE */ *background: #f00; /* IE 7 and below */ _background: #f60; /* IE 6 and below */  }', Rules.IEStarHack, Rules.IEStarHack);
+		// IE star hacks are incompatible with CSS nesting.
+		// assertRuleSet('selector { display: inline-block; *display: inline; }', Rules.IEStarHack);
+		// assertRuleSet('selector { background: #00f; /* all browsers including Mac IE */ *background: #f00; /* IE 7 and below */ _background: #f60; /* IE 6 and below */  }', Rules.IEStarHack, Rules.IEStarHack);
```
romainmenke, line 235: see : https://github.com/microsoft/vscode-css-languageservice/issues/344

IE star hacks aren't compatible with CSS Nesting. `*` triggers rule parsing.

src/test/css/nodes.test.ts
```diff
@@ -24,23 +24,35 @@ export function assertNodes(fn: (input: string) => nodes.Node, input: string, ex
...
+
+			while (actual.length > 0) {
+				if (nextExpectedSegment === nextActualSegment) {
+					break;
+				}
+
+				actualSegment = actual.shift()!;
+				nextActualSegment = actual[0];
+			}
+
```
romainmenke, line 47: This test helper was broken, it didn't actually test anything.
Adding any `...` in a expect string would cause the previous algorithm to just skip to the end in case of issues.

This new algorithm is a bit stricter, making it easier to write reliable tests for nested and repeating structures.

src/test/css/nodes.test.ts
```diff
@@ -102,9 +114,16 @@ suite('CSS - Nodes', () => {
...
 
+	function stylesheet(input: string): nodes.Stylesheet {
+		let parser = new Parser();
+		let node = parser.internalParse(input, parser._parseStylesheet)!;
+		return node;
+	}
+
 	test('RuleSet', function () {
-		assertNodes(ruleset, 'selector { prop: value }', 'ruleset,...,selector,...,declaration,...,property,...,expression');
-		assertNodes(ruleset, 'selector { prop; }', 'ruleset,...,selector,...,selector');
```
romainmenke, line 107: This was a false positive from the old test helper.
There is only one selector in this source, but the test still passed.

src/parser/cssParser.ts
```diff
@@ -361,16 +361,23 @@ export class Parser {
...
 			return this._parseRuleSetDeclarationAtStatement();
 		}
-		return (!this.peek(TokenType.Ident) && this._tryParseRuleset(true))
-			|| this._parseDeclaration();
+
+		if (!this.peek(TokenType.Ident)) {
+			return this._parseRuleset(true);
+		}
+
+		return this._tryParseRuleset(true) || this._parseDeclaration();
```
romainmenke, line 380: This still isn't fully correct for the CSS specification, but it does work ok in practice.

It is mainly important to have struct rule parsing when the current token isn't an ident. There must not a be a fallback to declaration parsing.

Doing the fallback to declaration parsing causes the wrong auto complete and diagnostics to be shown to users.

----

To be fully correct for the specification it would be better to do this :

1. parse as a declaration
   1. if this passes, return the declaration
   2. if this fails, record the diagnostics and reset the parser
2. parse as a rule
   1. if this passes, return the rule
   2. if this fails, record the diagnostics
3. report all diagnostics

It is unknown if a user is intending a declaration or a rule when both are invalid.

I think it is ok to add this later given that it is a very large refactor for an edge case.

romainmenke:
@aeschli friendly ping in case you missed these changes :)

aeschli:
@romainmenke Oh sorry, I missed that. PR look good, thanks a lot!

aeschli (approved):

joyceerhl (approved):

romainmenke:
Thank you @aeschli and @joyceerhl for reviewing this 🙇

PR #355, opened 2023-06-09
add support for `<timeline-range-name>` in `@keyframes`

romainmenke:
see : https://www.w3.org/TR/scroll-animations-1/#named-range-keyframes

With scroll driven animations, `@keyframes` rules can now look like this :

```css
@keyframes {
  cover 50% {
    color: purple;
  }
}
```

```
<keyframe-selector> = from | to | <percentage [0,100]> | <timeline-range-name> <percentage>
```


aeschli (approved):

aeschli:
Thanks a lot!

Tyriar (approved):

PR #365, opened 2023-10-05
support @container, prepare 6.2.10

aeschli:


aiday-mar (approved):

starball5:
what's the release milestone for this change?

aeschli:
It will be in 1.84

marcoms:
@aeschli Do I need to do something to enable support? I am on 1.86.2

Screenshot 2024-03-03 at 4 35 47 PM

aeschli:
@marcoms Are you using the rule nested? Can you open a new issue with a complete code example?

marcoms:
> @marcoms Are you using the rule nested? Can you open a new issue with a complete code example?

Yes this is nested, non-nested works fine. I noticed issue #381 was opened recently for SCSS so I added a code sample there, but if it makes more sense to tackle them separately I can make a new issue

PR #369, opened 2023-11-03
#74068 nth child specificity

SStranks:
Fix issue [#74068](https://github.com/microsoft/vscode/issues/74068) "[css] Wrong selector specificity values with :nth-child(even) and :nth-child(odd)". This patches the specificity for the nth-child/nth-last-child psuedo-selector. 

Debugging the code it seems the scanner/parser treats words 'even', 'odd', and 'n', as NodeType 16 (ElementNameSelector), which contributes to a false specificity. 

'div:nth-child(n of li)' should produce a specificity of 0-1-2, but produces 0-1-4 suggesting all three arguments are treated as elements. If the 'n' has an integer prefix the correct specificity is produced however.

Rather than try to unpack the scanner/parser logic, the solution exploits the formula of the psuedo-selector by testing for the presence of " of " within the arguments string, ["An+B [of S]?"](https://www.w3.org/TR/selectors-4/#the-nth-child-pseudo). 

> The specificity of the :nth-child(An+B [of S]?) pseudo-class is the specificity of a single pseudo-class plus, if S is specified, the specificity of the most specific complex selector in S

### DIFF
--- a/src/services/selectorPrinting.ts
+++ b/src/services/selectorPrinting.ts
@@ -456,10 +456,9 @@ export class SelectorPrinting {
 							continue elementLoop;
 						}
 
-						if (text.match(/^:(?:nth-child|nth-last-child|host|host-context)/i) && childElements.length > 0) {
+						if (text.match(/^:(?:host|host-context)/i) && childElements.length > 0) {
 							// The specificity of :host() is that of a pseudo-class, plus the specificity of its argument.
 							// The specificity of :host-context() is that of a pseudo-class, plus the specificity of its argument.
-							// The specificity of an :nth-child() or :nth-last-child() selector is the specificity of the pseudo class itself (counting as one pseudo-class selector) plus the specificity of the most specific complex selector in its selector list argument.
 							specificity.attr++;
 
 							let mostSpecificListItem = calculateMostSpecificListItem(childElements);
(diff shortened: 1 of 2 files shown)


nicolasploquin (commented):
Code review only, no build to test.
Good job 👍

src/services/selectorPrinting.ts
```diff
@@ -470,6 +469,27 @@ export class SelectorPrinting {
...
 
+						if (text.match(/^:(?:nth-child|nth-last-child)/i) && childElements.length > 0) {
+							/* The specificity of the :nth-child(An+B [of S]?) pseudo-class is the specificity of a single pseudo-class plus, if S is specified, the specificity of the most specific complex selector in S */
+							// https://www.w3.org/TR/selectors-4/#the-nth-child-pseudo
+							specificity.attr++;
+
+							const selectorText = text.slice(text.indexOf('(') + 1, text.length -1);
+
+							// Test for presence of complex-selector-list S; find highest specificity
+							const keyword = text.indexOf(' of ');
```
aeschli, line 480: Testing for a space before and after `of` is not clean. There could also be a tab, or even a non word character
```
ul.one > li:nth-child(-n + 3 of.noted) {
}
```
would be valid too.

I suggest to jeust use the scanner to find out if there's an `of` ident in the string.

aeschli (changes requested):

src/services/selectorPrinting.ts
SStranks: Reworked the logic in the latest commit. When the 'An + B [of S]?' syntax is respected then we can just grab the last childElement as it will contain the entire complex-selector list (S). 

The edge case where 'n' has no integer prefix (the parser does not recognise it as a binary expression token). Extract out the complex-selector list, and re-parse the selectors. This also handles the different syntax possibilities; added more tests for completeness.

aeschli:
Looks good! If you can run the fomatter on `selectorPrinting.ts` one more time that would be perfect.

SStranks:
> Looks good! If you can run the fomatter on `selectorPrinting.ts` one more time that would be perfect.

Amended

aeschli:
Thanks a lot!

alexr00 (approved):

PR #373, opened 2023-11-18
[scss] Path resolver to include partial files support

dyhagho:
## Context

When importing [partials](https://sass-lang.com/documentation/at-rules/import/#partials) (e.g "_mixins.scss"), it's unnecessary to explicitly include the underscore (`_`). Sass inherently understands and imports files with or without the underscore. Therefore, `@import "mixins.scss"` is a valid way to import the partial.

## Why this change has been made?

Navigating to a partial file using `Command + Click` or `Control + Click` without specifying the underscore was causing the language server to be unable to find the path.

## Changes
- Update  `toPathVariations` function to consider partial files when the underscore has been omitted.
-  Add a new test case to ensure accurate resolution when the underscore is omitted when import partial files.

### Before
before

### After
after

### DIFF
--- a/src/services/scssNavigation.ts
+++ b/src/services/scssNavigation.ts
@@ -44,8 +44,8 @@ export class SCSSNavigation extends CSSNavigation {
 }
 
 function toPathVariations(target: string): DocumentUri[] {
-	// No variation for links that ends with suffix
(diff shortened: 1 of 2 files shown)


dyhagho:

aeschli (approved):

justschen (approved):

PR #384, opened 2024-02-16
feat: add support for Sass `pkg:` importers

### LINKED ISSUE #383: Support Sass Node package importer
As part of [Sass 1.71](https://github.com/sass/dart-sass/releases/tag/1.71.0), a new concept of "package importer" has been introduced, allowing authors to write `@use "pkg:bootstrap"` instead of `@use "~bootstrap"` or `@use "../../node_modules/bootstrap"`. Sass 1.71 includes a built-in package importer that supports Node modules.

**Current behavior**
CMD Clicking on a `pkg:` URL in a Sass file shows a "Unable to open 'pkg:..." modal, as the language service doesn't know how to resolve the files.

**Expected behavior**
CMD Clicking on a `pkg:` URL would take you to the corresponding file.

The resolution is almost identical to `~`, resolving partials and index files. In addition, it also resolves [conditional exports](https://nodejs.org/api/packages.html#conditional-exports), and resolves in the following order:

1. sass, style, or default condition in package.json exports.
2. If there is not a subpath, then find the root export:
   1. sass key at package.json root.
   2. style key at package.json root.
   3. index file at package root, resolved for file extensions and partials.
3. If there is a subpath, resolve that path relative to the package root, and resolve for file extensions and partials.

[Blog post](https://sass-lang.com/blog/announcing-pkg-importers/)
[Spec](https://github.com/sass/sass/blob/main/accepted/package-importer.d.ts.md)

wkillerud:
Fixes #383

Hello 👋 I'm opening a PR with a suggested solution for #383

As mentioned in the issue, Sass released a [new syntax for import strings](https://sass-lang.com/blog/announcing-pkg-importers/), which I imagine trips up a lot of tools. I find the documentation for [NodePackageImporter](https://github.com/sass/sass/blob/f6832f974c61e35c42ff08b3640ff155071a02dd/js-api-doc/importer.d.ts#L349) has the most readable explanation of the different supported scenarios. Also relevant here is the Node documentation on [subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns).

It's a big chunk with many different code paths, I'll admit. I've tried documenting at a reasonable level, and added tests that should cover a lot of scenarios. That said, I understand if you're hesitant to add yet more module resolution logic here.

I've tested locally with `npm link` in a [running extension using these test fixtures](https://github.com/wkillerud/vscode-scss/commit/86e7be3c73847fd7a983c0f69837124eb5bb9407), and the resolution provides correct links for me. Recording of that below.

https://github.com/microsoft/vscode-css-languageservice/assets/1223410/37020935-38a8-4706-9c36-b07a34e7f166

## Considered alternatives

Sass's JavaScript API includes the aforementioned [NodePackageImporter](https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/). Its implementation seems to have a function [`canonicalize`](https://github.com/sass/dart-sass/blob/3e6721e79f049dd01880542667380640c2d1eeae/lib/src/importer/node_package.dart#L37), which looked promising as a method of looking up an import string. Unfortunately it's not part of the public API.

The result from `sass.compile` includes all the loaded URLs, so I'm sure you could hack something together there (pass in a string that only `@use`s the `pkg:module` import in question). It would require bundling `sass` though, which is also [technically possible](https://sass-lang.com/blog/sass-in-the-browser/) now.

### DIFF
--- a/src/cssLanguageTypes.ts
+++ b/src/cssLanguageTypes.ts
@@ -154,7 +154,7 @@ export namespace ClientCapabilities {
 
 export interface LanguageServiceOptions {
 	/**
-	 * Unless set to false, the default CSS data provider will be used 
+	 * Unless set to false, the default CSS data provider will be used
 	 * along with the providers from customDataProviders.
 	 * Defaults to true.
 	 */
@@ -284,7 +284,7 @@ export interface FileStat {
 export interface FileSystemProvider {
 	stat(uri: DocumentUri): Promise<FileStat>;
 	readDirectory?(uri: DocumentUri): Promise<[string, FileType][]>;
-	getContent?(uri: DocumentUri, encoding?: BufferEncoding): Promise<string>;
+	getContent?(uri: DocumentUri, encoding?: string): Promise<string>;
 }
 
 export interface CSSFormatConfiguration {
@@ -306,11 +306,11 @@ export interface CSSFormatConfiguration {
 	preserveNewLines?: boolean;
 	/** maximum number of line breaks to be preserved in one chunk. Default: unlimited */
 	maxPreserveNewLines?: number;
-    /** maximum amount of characters per line (0/undefined = disabled). Default: disabled. */
+	/** maximum amount of characters per line (0/undefined = disabled). Default: disabled. */
 	wrapLineLength?: number;
 	/** add indenting whitespace to empty lines. Default: false */
 	indentEmptyLines?: boolean;
-	
+
 	/** @deprecated Use newlineBetweenSelectors instead*/
 	selectorSeparatorNewline?: boolean;
 
--- a/src/cssLanguageTypes.ts
+++ b/src/cssLanguageTypes.ts
@@ -284,6 +284,7 @@ export interface FileStat {
 export interface FileSystemProvider {
 	stat(uri: DocumentUri): Promise<FileStat>;
 	readDirectory?(uri: DocumentUri): Promise<[string, FileType][]>;
+	getContent?(uri: DocumentUri, encoding?: BufferEncoding): Promise<string>;
 }
 
 export interface CSSFormatConfiguration {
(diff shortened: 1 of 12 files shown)


src/cssLanguageTypes.ts
```diff
@@ -284,6 +284,7 @@ export interface FileStat {
 export interface FileSystemProvider {
 	stat(uri: DocumentUri): Promise<FileStat>;
 	readDirectory?(uri: DocumentUri): Promise<[string, FileType][]>;
+	getContent?(uri: DocumentUri, encoding?: BufferEncoding): Promise<string>;
```
wkillerud, line 287: Matches the [`RequestService`](https://github.com/microsoft/vscode/blob/6077d59e422b88df1ec5ce7b539bcf50dd7077f6/extensions/css-language-features/server/src/cssServer.ts#L65) on the language server side

src/services/cssNavigation.ts
```diff
@@ -392,6 +392,92 @@ export class CSSNavigation {
 			return this.mapReference(await this.resolveModuleReference(target, documentUri, documentContext), isRawLink);
 		}
 
+		// Following the [sass package importer](https://github.com/sass/sass/blob/f6832f974c61e35c42ff08b3640ff155071a02dd/js-api-doc/importer.d.ts#L349),
+		// look for the `exports` field of the module and any `sass`, `style` or `default` that matches the import.
+		// If it's only `pkg:module`, also look for `sass` and `style` on the root of package.json.
+		if (target.startsWith('pkg:')) {
+			const bareTarget = target.replace('pkg:', '');
```
aeschli, line 399: can this be extracted in a method ?

src/services/cssNavigation.ts
```diff
@@ -460,6 +546,18 @@ export class CSSNavigation {
...
 	}
 
+	protected async getContent(uri: string): Promise<string | null> {
+		if (!this.fileSystemProvider || !this.fileSystemProvider.getContent) {
+			return null;
+		}
+		try {
+			return await this.fileSystemProvider.getContent(uri);
+		} catch (err) {
+			console.error(err);
```
aeschli, line 556: Should not print to console

aeschli (changes requested):

Asmodeusxxx85 (approved):

aeschli (approved):

aeschli:
Thanks @wkillerud. I made some improvements and moved the code to `scssNavigation`.

chrmarti (approved):

PR #390, opened 2024-03-30
support nested @container

liaoyinglong:
Close support nested @container  #389

PR #421, opened 2025-01-01
Support `@starting-style`

### LINKED ISSUE #403: Support `@starting-style`
Using the relatively new `@starting-style` yields a warning:

image

Docs:
https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style

Spec:
https://drafts.csswg.org/css-transitions-2/#defining-before-change-style-the-starting-style-rule

CanIUse: https://caniuse.com/mdn-css_at-rules_starting-style (every major browser supports it)

lilnasy:
- Closes #403
- I implemented this by following the lead of #365 and #390. If there is more to it, then lmk.
- I noticed that `._parseContainerDeclaration()`, `._parseMediaDeclaration()`, `._parseSupportsDeclaration()`, and `._parseLayerDeclaration()` all contain the exact same logic. It seems community contributions over the years kept copying it.
- To keep the review simple and the diff small, I didn't break the pattern and created a 5th copy in `._parseStartingStyleDeclaration()`. Might be something to revisit.

### DIFF
--- a/src/parser/cssNodes.ts
+++ b/src/parser/cssNodes.ts
@@ -101,8 +101,8 @@ export enum NodeType {
 	PropertyAtRule,
 	Container,
 	ModuleConfig,
+	SelectorList,
 	StartingStyleAtRule,
-	SelectorList
 }
 
 export enum ReferenceType {
--- a/src/parser/cssNodes.ts
+++ b/src/parser/cssNodes.ts
@@ -101,6 +101,7 @@ export enum NodeType {
 	PropertyAtRule,
 	Container,
 	ModuleConfig,
+	StartingStyleAtRule,
 }
 
 export enum ReferenceType {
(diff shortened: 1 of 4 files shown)


lilnasy:

lilnasy:
@aeschli The review comments were addressed.

src/parser/cssNodes.ts
```diff
@@ -101,6 +101,7 @@ export enum NodeType {
 	PropertyAtRule,
 	Container,
 	ModuleConfig,
+	StartingStyleAtRule,
 	SelectorList
```
lilnasy, line 105:
```suggestion
	SelectorList,
	StartingStyleAtRule,
```

aeschli:
Thanks @lilnasy, looks good!

aeschli (approved):

rebornix (approved):

aeschli:
@aarontgrogg The latest VS Code release is 1.108

aarontgrogg:
> @aarontgrogg The latest VS Code release is 1.108

Crud, Update Mode was set to "Never" for some reason... 
Odd that that makes the "Check for Updates" menu option disappear completely, but I guess that makes sense?
Anyway, thanks, setting changed, update installed, error message gone!
Cheers! 🍻

Issue #427, opened 2025-03-17
Add Baseline status to CSS hovercards


rviscomi:
[Baseline](https://web-platform-dx.github.io/web-features/) is an initiative by the W3C WebDX Community Group that aims to improve how we communicate the level of browser compatibility of web platform features.

In VS Code today, CSS properties have a hovercard that shows supported browser versions:

![Image](https://github.com/user-attachments/assets/69e8ba89-5854-4035-9540-7870ca9daba4)

Just by looking at this information, it's not immediately clear if this feature is ready to use in production. This could be improved by adopting a similar treatment as the one used by [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit):

![Image](https://github.com/user-attachments/assets/23a601ad-3617-46f7-9ca2-2096c02377d5)

Here, the Baseline "widely available" logo and text are used to more clearly communicate that the feature has been available across the core browser set for a substantial amount of time (2.5 years).

![Image](https://github.com/user-attachments/assets/7235473e-0a3e-4bd8-88cf-18dcac8b0640)

To see how this might look, I prototyped a VS Code [extension](https://github.com/rviscomi/baseline-vscode) that displays a hovercard with a similar treatment.

After a quick chat with @aeschli, it sounds like this could be a useful addition. Filing this issue to continue the discussion.

aeschli:
Maybe we can use`compute-basline` in vscode-custom-data and add a new property to the generated data.

I try to update `vscode-custom-data` regularly

Lets start with CSS properties

rviscomi:
@aeschli would you like me to start looking into adding support to HTML hovercards too?

aeschli:
Is there a need for compatibility information with HTML? Sorry my ignorance, but are there elements/properties that are not supported/deprecated...?

rviscomi:
Yeah good question. There are some newer HTML elements that aren't widely available yet, like [`<fencedframe>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fencedframe). Even for well established elements like `<script>` and `<img>`, new attributes and attribute values can get added over time. For example [`<img fetchpriority>`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority) and [`<script type="speculationrules">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/speculationrules) aren't widely available either.

rviscomi:
[...]

PR #428, opened 2025-03-18
Add Baseline status to hovercards

rviscomi:
Progress on #427

- [x] sync custom-data dependency with the output of https://github.com/microsoft/vscode-custom-data/pull/111
- [x] add Baseline info in hovercards
- [x] update tests

Hovercard changes:
- adds baseline icon and status
  - replaces the browser version list
  - icon supports light/dark mode and corresponds to the feature status
- omits the experimental warning (covered by limited availability status)

Demo:

https://github.com/user-attachments/assets/8cf066b2-2e09-4c95-9579-51423cab7f12

<details>

<summary>Screenshots...</summary>

### Widely available

Before:
image

After (dark mode):
image

After (light mode):
image

### Newly available

Before:
image

After (dark mode):
image

After (light mode):
image

### Limited availability

Before:
image

After (dark mode):
image

After (light mode):
image

### Experimental

Before:
image

After (dark mode):
image

After (light mode):
image

(experimental warning removed, as it's redundant with the limited availability message)

### Obsolete

Before:
image

After (dark mode):
image

After (light mode): 
image

(obsolete warning kept)

</details>


rviscomi:
FYI I was getting a TypeScript error related to my use of the `Intl.ListFormat` function:

```
$ npm run test

> vscode-css-languageservice@6.3.2 test
> npm run compile && npm run mocha


> vscode-css-languageservice@6.3.2 compile
> tsc -p ./src && npm run copy-jsbeautify && npm run lint

src/languageFacts/entry.ts:193:50 - error TS2339: Property 'ListFormat' does not exist on type 'typeof Intl'.

193 const missingBaselineBrowserFormatter = new Intl.ListFormat("en", {
                                                     ~~~~~~~~~~


Found 1 error in src/languageFacts/entry.ts:193
```

I can fix it by upgrading the tsconfig to es2021, but figured that would need to be part of a separate discussion. For now I worked around it by changing it to `(Intl as any)`.

src/languageFacts/entry.ts
```diff
@@ -141,32 +196,34 @@ function getEntryMarkdownDescription(entry: IEntry2, settings?: HoverSettings):
...
+	if (!support) {
+		return '';
+	}
+	const missingBrowsers = new Map(Object.entries(browserNames));
+	for (const browser in support) {
+		missingBrowsers.delete(browser);
+	}
+
+	return missingBaselineBrowserFormatter.format(Object.values(Array.from(missingBrowsers.entries()).reduce((browsers: Record<string, string>, [browserId, browser]) => {
+		if (browser.name in browsers || browserId == 'edge') {
```
captainbrosset, line 219: The core browser set is not likely to change much, but if it does, this hard cording (as well as the `browserNames` const are a bit unfortunate as they'll need to be changed if the core browser set changes.
Can you think of a utility package/function which we could maintain on the web-features repo instead, which would avoid consumers like this one to have to do this?

rviscomi: Hm actually the `computeBaseline` function does already expose browser names and platforms, eg:

```
Browser {
    id: 'chrome',
    data: {
      accepts_flags: true,
      accepts_webextensions: true,
      name: 'Chrome',
      pref_url: 'chrome://flags',
      preview_name: 'Canary',
      releases: [Object],
      type: 'desktop'
    },
    ...
}
```

I'll look into moving the "missing browser" functionality over to [vscode-custom-data#111](https://github.com/microsoft/vscode-custom-data/pull/111/files#diff-df39d64e75bdd56b21124ae2fcdc9778ae695292a82c50630bd1ebfdbc9e9f2aR374) where it already uses `computeBaseline`.

rviscomi: Given @aeschli's [comment](https://github.com/microsoft/vscode-custom-data/pull/111#issuecomment-2740307413) about the need for a well-defined schema, I think this is actually ok. If the core browser set changes, this implementation will need to be updated in a few places anyway.

aeschli:
````
      "browsers": [
        "E12",
        "S10",
        "C50",
        "IE10",
        "O37"
      ],
      "baseline": {
        "status": "high",
        "support": {
          "chrome": "50",
          "chrome_android": "50",
          "edge": "12",
          "firefox": "28",
          "firefox_android": "28",
          "safari": "10",
          "safari_ios": "10"
        },
        "baseline_low_date": "2015-09-30",
        "baseline_high_date": "2018-03-30"
      },
 ```


Now the browser information is duplicated, or have conflicting information? Can't we use the browsers we already have? 
It's ok to add new browsers (just need to spec it)
If it's incomplete or has old data, we can set the data based on the baseline call.

**Comment by rviscomi** [comment]
@aeschli sure I've removed the `baseline.support` object and added the mobile versions of the Baseline core browser set to the existing `browsers` array

**Comment by rviscomi** [comment]
Found an issue with the [`accent-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) property, which is not Baseline due to an implementation issue in Safari. In BCD, this is indicated by the [`partial_implementation`](https://github.com/mdn/browser-compat-data/blob/c6992676d95f20b9d9be3a06726cd609e8a2800d/css/properties/accent-color.json#L27-L31) flag. `compute-baseline` takes this into consideration and omits Safari from the list of supported browsers. But Safari still technically has a supported version in BCD and ends up in the list.

Actual:

image

Expected:

image

I fixed this by removing support strings from the `browsers` array for any core browsers missing from Baseline's support list.

**Comment by aeschli** [comment]
Thanks for the PR. I think we need to improve the look of the hover.
- I don't find the baseline icon intuitive. I would would go with the ⚠️ and a :heavy_check_mark: 
- I would not put the name first

My suggestion:
- description
- compatibility warning (Property is experimental. Property is obsolete, Property is nonstandard)
- baseline
- syntax

**Comment by aeschli** [comment]
The mockups are good. I still don't like the extra wide baseline icons. It's a bit too large for the text size, also I find the form not very intuitive. Is the shape important to you? Can we not just add a checkmark?

**Comment by aeschli** [comment]
BTW I adopted the custom data, but didn't make any changes to the hover. Leaving this to you.
https://github.com/microsoft/vscode-css-languageservice/pull/431/files

**Comment by rviscomi** [comment]
Here's an updated mockup of a 10px tall icon instead of the 14px version in light mode:

image

If that looks good to you, I'll push a commit updating the rest of the icons.

**Comment by captainbrosset** [comment]
Regarding the use of icons, thank you @aeschli for your feedback on the fact that they're not intuitive. @atopal, let's add this to other feedback we might have received and reflect on whether improvements are needed on the long run.

For now, though (i.e., for this PR), I strongly suggest keeping the Baseline icons as they are defined now, and not deviate from them.
Baseline is a concept that was introduced a couple of years ago and is getting adopted by more and more sites, dashboards, and tools. We pay a lot of attention as to how Baseline is being adopted, and sometimes even work hand in hand with consumers to make sure that Baseline is communicated correctly.
The reason is that Baseline is an attempt at summarizing the many details of a very fragmented and nuanced web platform, into something that's as simple as 3 statuses only. This is essentially impossible to do, but we believe that doing it in the way we're doing it now still has more benefits than drawbacks. If, and only if, Baseline is communicated correctly.

Changing the icons to a checkmark and a warning icon would simplify what Baseline is even more, to the point that it becomes dangerous.
Indeed, Baseline says "if this is widely available, then you don't really need to worry about it, or test much". That's basically all it says.
Baseline **does not** say "if this is not widely available, then don't use it". If a feature is not widely available, then it's basically up to the developer to make  conscious decision as to whether they want to use that feature. They may very well use it, even if the feature has limited availability, if they've tested it for their specific use case. Maybe they're using a polyfill in other browsers. Maybe it's part of a progressively enhanced feature set. Who knows.

So, Baseline messages/icons should never appear to developers as something that may steer them away from using a feature.

**Comment by aeschli** [comment]
The smaller baseline icon looks better. Lets go with that!

**Review by aeschli — APPROVED** [review]

**Comment by aeschli** [comment]
@rviscomi Thanks Rick, nice work, it looks great!

**Review by alexr00 — APPROVED** [review]

**Comment by rviscomi** [comment]
Thanks for all your help and can't wait to see it live! 🥳

File: src/languageFacts/entry.ts [hunk 2]
```diff
@@ -40,6 +56,39 @@ function getEntryStatus(status: EntryStatus) {
...
+	if (baseline.status === "false") {
+		const missingBrowsers = getMissingBaselineBrowsers(browsers);
+		let status = `Limited availability across major browsers`;
+		if (missingBrowsers) {
+			status += ` (Not fully implemented in ${missingBrowsers})`;
+		}
+		return status;
+	}
+
+	const baselineYear = baseline.baseline_low_date?.split('-')[0];
```
FluorescentHallucinogen, line 69: @rviscomi If the status is `low`, then `baseline_low_date` is used, but if the status is `high`, then again `baseline_low_date` is used, not `baseline_high_date`. It turns out that `baseline_high_date` is not used at all. Is this not a bug? Is this by design?

rviscomi: Hi, yes this is by design and is consistent with other Baseline resources.  For example, the [`dialog`](https://github.com/web-platform-dx/web-features/blob/2898cf81bb72673ecf32d04089d8fa312e6287ad/features/dialog.yml.dist#L1-L7) element is widely available since 2024, and Baseline since 2022:

```
status:
  baseline: high
  baseline_low_date: 2022-03-14
  baseline_high_date: 2024-09-14
```

Resources like [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) and the [Web Platform Dashboard](https://webstatus.dev/features/dialog) show the newly available date:

image

image

src/languageFacts/entry.ts
```diff
@@ -141,31 +196,39 @@ function getEntryMarkdownDescription(entry: IEntry2, settings?: HoverSettings):
 	return result;
 }
 
+// TODO: Remove "as any" when tsconfig supports es2021+
+const missingBaselineBrowserFormatter = new (Intl as any).ListFormat("en", {
+	style: "long",
+	type: "disjunction",
```
FluorescentHallucinogen, line 202: @rviscomi Why did you use disjunction instead of conjunction here? 🤔

rviscomi: "or" is more correct. Here's Gemini to help articulate exactly why that is 😄 

> When listing items that share a common negative characteristic, using "or" is generally more grammatically precise and less ambiguous.
>
>So, the more correct phrasing is:
>**"Not fully implemented in Chrome, Edge, or Safari"**
>
>Here's why:
>
>1. Meaning with "or": This sentence structure implies that the negation ("Not fully implemented") applies to each browser individually:
>    - It is not fully implemented in Chrome,
>    - AND it is not fully implemented in Edge,
>    - AND it is not fully implemented in Safari. This accurately conveys that none of the listed browsers have full implementation. In logic, this corresponds to `¬A ∧ ¬B ∧ ¬C`.
>2. Potential ambiguity with "and": The sentence "Not fully implemented in Chrome, Edge, and Safari" could be interpreted to mean: "It is not the case that the feature is fully implemented in all three (Chrome, Edge, and Safari) simultaneously."
>    - This statement would be true even if the feature was implemented in Chrome and Edge but not in Safari (i.e., at least one of them doesn't have it). In logic, this corresponds to `¬(A ∧ B ∧ C)`, which is equivalent to `¬A ∨ ¬B ∨ ¬C` (at least one is not implemented).
>    - While readers might often infer the intended meaning (that none of them have it), the "and" phrasing is technically ambiguous and not as precise for indicating that each listed item lacks the feature.
>For clear technical documentation, where precision is important, "or" is the preferred conjunction in this negative context to ensure the meaning is that each of the browsers listed does not support the feature.

src/languageFacts/entry.ts
```diff
@@ -4,33 +4,49 @@
...
+		name: 'Safari',
+		platform: 'macOS'
+	},
+	'SM': {
+		name: 'Safari',
+		platform: 'iOS'
+	}
 };
 
+const shortCompatPattern = /(E|FFA|FF|SM|S|CA|C|IE|O)([\d|\.]+)?/;
```
FluorescentHallucinogen, line 40: @rviscomi Why are we looking for the `|` character in the `[\d|\.]` part? 🤔

I've analyzed the `browsers.css-data.json` file from the `@vscode/web-custom-data` package and haven't found a single property that has this character in the `browsers` or  `values.browsers` field.

But I've found 5 properties that have values in the `browsers` field, that don't match current pattern:

`overflow-anchor`: `'Spreview'`
`-moz-background-clip`: `'FF1-3.6'`
`-ms-filter`: `'IE8-9'`
`-ms-transform`: `'IE9-9'`
`-ms-transform-origin`: `'IE9-9'`

Actually, we are not interested in the version value (the second part of the regexp), only the browser name (the first part of the regexp).

Maybe we can do with a simpler regexp or without using a regexp at all?

aeschli: You can remove the `|`...

PR #438, opened 2025-04-24
refactor: Extend the base tsconfig.json

rgant:
__DESCRIPTION:__
I noticed that both the `tsconfig.json` and the `tsconfig.esm.json` files are largely the same. So it makes sense to me to extend the base on instead of duplicating the configuration options.


aeschli (approved):

aeschli:
Thanks @rgant !

benibenj (approved):

PR #440, opened 2025-04-26
SCSS: resolve bare module imports on Windows (#245579)

IsaacTian05:
Fix issue https://github.com/microsoft/vscode-css-languageservice/issues/432


src/test/scss/scssNavigation-node-modules.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const ls = getSCSSLanguageService();
+const mockFS = getNodeFSRequestService();
+
+describe('SCSS link navigation – node_modules', () => {
+  it('resolves bootstrap path on Windows', async () => {
+    const doc = TextDocument.create('file:///c:/proj/app.scss', 'scss', 1,
+        "@import 'bootstrap/scss/variables';");
+    const links = await ls.findDocumentLinks2(doc, ls.parseStylesheet(doc), {}, mockFS);
+    const target = links[0].target!.replace(/\\/g, '/');
```
aeschli, line 14: target is a URI, it should not contain `\`

aeschli (changes requested):

triggerbuild.txt
```diff
@@ -0,0 +1 @@
+build
```
aeschli, line 1: This should also be removed....

src/services/cssNavigation.ts
```diff
@@ -422,7 +422,20 @@ export class CSSNavigation {
 			return this.mapReference(await this.resolveModuleReference(target, documentUri, documentContext), isRawLink);
 		}
 
-		const ref = await this.mapReference(documentContext.resolveReference(target, documentUri), isRawLink);
+		// Treat bare module names (“bootstrap/...”) like sass-loader does
+		const isBareImport = !target.startsWith('.')             // not ./ or ../
+		                 && !target.startsWith('/')              // not workspace-absolute
+		                 && target.indexOf(':') === -1;          // not a scheme (file://)
```
aeschli, line 428: Maybe a use a regex to make that check [
startsWithSchemeRegex = /^\w[\w\d+.-]:/

aeschli (changes requested):
Looks good. If you can add some more test cases that would be great

src/services/cssNavigation.ts
```diff
@@ -422,7 +422,22 @@ export class CSSNavigation {
 			return this.mapReference(await this.resolveModuleReference(target, documentUri, documentContext), isRawLink);
 		}
 
-		const ref = await this.mapReference(documentContext.resolveReference(target, documentUri), isRawLink);
+		// Treat bare module names (“bootstrap/...”) like sass-loader does
+		const startsWithSchemeRegex = /^\w[\w\d+.-]:/;
```
aeschli, line 426: Node module resolving should only happen when `this.resolveModuleReferences` is true.
(only set for LESS and SCSS).

So this needs to go behind the `if (this.resolveModuleReferences) {`.
I looks like we already call `resolveModuleReference` there, so this code is not necessary?

aeschli (changes requested):

IsaacTian05 (commented):
Pending

miss-programgamer:
If these changes are still pending, I can go ahead and implement them.

...Actually, how *would* I go about adding changes to someone else's pull request?

aeschli:
replaced by #459

PR #256, opened 2021-12-06
feature: autocomplete for media queries

jesperjen98:
Implementation of media query autocomplete for both media types and
features. The implementation is based on how the IProperyData interface
is implemented and may therefor contain a small amount of redundant
code. The modified files are: cssLanguageTypes.ts, webCustomData.ts,
dataManager.ts, dataprovider.ts, entry.ts, cssCompletion.ts.

PR #443, opened 2025-05-29
Add basic media query auto complete support

Legend-Master:
Waiting for https://github.com/microsoft/vscode-custom-data/pull/118

Closes #256

### DIFF
--- a/src/cssLanguageTypes.ts
+++ b/src/cssLanguageTypes.ts
@@ -198,6 +198,16 @@ export interface IPropertyData {
 	relevance?: number;
 	atRule?: string;
 }
+export interface IDescriptorData {
+	name: string,
+	description?: string,
+	references: IReference[],
+	syntax: string,
+	type: string,
+	values?: IValueData[]
+	browsers?: string[];
+	baseline?: BaselineStatus;
+}
 export interface IAtDirectiveData {
 	name: string;
 	description?: string | MarkupContent;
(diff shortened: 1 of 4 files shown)


aeschli:
Nice, thanks a lot!

Tyriar (approved):

Tyriar (approved):

germanfrelo:
Great addition! Although I can't see any autocomplete suggestions for `@media (min-` or `@media (max-)`:

Image

Is that intentional?

I'm using Visual Studio Code (all extensions disabled):

- Version: 1.101.0
- Commit: dfaf44141ea9deb3b4096f7cd6d24e00c147a4b1
- Date: 2025-06-11T15:00:50.123Z
- Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) - Code/1.101.0 Chrome/134.0.6998.205 Electron/35.5.1 Safari/537.36

Legend-Master:
This syntax is derived from the [range type](https://drafts.csswg.org/mediaqueries-5/#mq-ranges) and needs some special handling, and that is not implemented yet

To be honest though it's kinda intentional to me, since I find that syntax quite hard to reason about, and most of the time, if you use anything like postcss (included in many bundlers by default), the new syntax (`@media (width <= 300px) {}`) would be transformed automatically if you target old browsers, also [this is baseline 2023 so most browsers already support them natively](https://web.dev/articles/media-query-range-syntax)

germanfrelo:
Fair enough.

PR #459, opened 2025-09-30
SCSS: resolve bare module imports on Windows (the sequel)

miss-programgamer:
Duplicate of https://github.com/microsoft/vscode-css-languageservice/pull/440 because original user went awol.


miss-programgamer:

miss-programgamer:
@aeschli

aeschli:
## Changes Summary

PR #459 adds bare SCSS module resolution on Windows, including root-level module imports, while preserving scheme-prefixed URLs.

## What’s Done Well

- Handles imports that reference only a module root.
- Includes coverage for standard and custom URI schemes.
- Compilation and linting pass.

## Critical Issues

1. **Correctness — relative imports now lose precedence**  
   `cssNavigation.ts:443` tries `node_modules` before checking whether the relative target exists. This breaks existing behavior when both a local directory and package have the same name.  
   Confirmed failures:
   - CSS `green/c.css` resolves to `node_modules/green/c.css`
   - SCSS `green/d.scss` resolves to `node_modules/green/d.scss`

2. **Tests — new Windows test fails on macOS/Linux**  
   `scssNavigation.test.ts:410` references a nonexistent `c:/proj/node_modules/bootstrap` fixture and constructs a platform-dependent expected URI. The test falls back to `file:///c:/proj/bootstrap/scss/variables` instead of resolving the package.

## Improvements

- The scheme tests beginning at `scssNavigation.test.ts:374` do not prove the URLs avoided bare-import classification: module lookup can fail and still return the expected unchanged URL.

## Validation

```text
npm run compile
node --enable-source-maps --test lib/esm/test/scss/scssNavigation.test.js
```

Compilation and linting passed; tests reported **45 passed, 3 failed**.

## Verdict

**Needs Revisions.** Restore relative-file-first resolution, then make the Windows test use a real or mocked package fixture with platform-independent URI assertions.

src/services/cssNavigation.ts
```diff
@@ -439,7 +439,18 @@ export class CSSNavigation {
 			return this.mapReference(await this.resolveModuleReference(target, documentUri, documentContext), isRawLink);
 		}
 
-		const ref = await this.mapReference(documentContext.resolveReference(target, documentUri), isRawLink);
+		// Treat bare module names (“bootstrap/...”) like sass-loader does
+		if (this.resolveModuleReferences && importIsBare(target)) {
+			const resolvedModuleRef = await this.resolveModuleReference(target, documentUri, documentContext);
+			const moduleRef = await this.mapReference(resolvedModuleRef, isRawLink);
+
+			if (moduleRef != null) {
```
aeschli, line 447: Should be `moduleRef !== undefined`

aeschli (changes requested):
See the review comments: https://github.com/microsoft/vscode-css-languageservice/pull/459#issuecomment-5583363640

PR #460, opened 2025-10-14
#163967 css alias

SStranks:
In conjunction with Pull Request [#197033](https://github.com/microsoft/vscode/pull/197033) on the vscode repository.

I reworked the previous merged code for this feature so that it now handles single-root, multi-root, and mono-repo project structures. 

To clarify, aliases for css/scss are specified in the `.vscode/settings.json` under 'importAliases' for each language as before - I have updated the tool-tip descriptions in the package.nls.json file, and made code changes to cssServer to accommodate the functionality in the current PR - see PR 197033 above.

### Functionality

- Settings files/aliases at the workspace root take priority over any nested settings files. This allows avoiding directory tree lookup for settings files if a valid alias is discovered there first.
- If no settings/alias is found at the workspace root; attempt lookup for `.vscode/settings.json` from document towards workspace root. 
- If lookup finds settings file; cache the path and cache the aliases for that document to avoid repeat lookups.
- Added Public `clearCache()` method to `CSSNavigation`; to be called when settings.json changes (see limitations section below)

### Changes

- Added functionality above.
- Updated: navigation tests.
- Added: test fixture files

### Limitations

- `clearCache()` method only fires when settings files tracked by vscode are changed. Settings files discovered by lookup in cssNavigation are not covered. I didn't see a straightforward way to implement this; vscode cssServer imports `connection.onDidChangeConfiguration` from vscode-languageserver, and handles the watcher. Willing to make further changes if someone can suggest how/where to look.


src/services/cssNavigation.ts
```diff
@@ -22,17 +23,28 @@ import {
 } from '../languageFacts/facts.js';
 import { startsWith } from '../utils/strings.js';
 import { dirname, joinPath } from '../utils/resources.js';
+import { readFile } from 'node:fs/promises';
```
aeschli: We must not use nodejs functionality as the code can also run in the browser.

To read files use the fileSystemProvider

aeschli (changes requested):

src/services/cssNavigation.ts
```diff
@@ -458,27 +475,109 @@ export class CSSNavigation {
...
+	}
+
+	private async parseSettingsFile(settingsPath: string): Promise<VSCodeSettings | undefined> {
+		const candidate = URI.parse(settingsPath);
+		try {
+			const text = await readFile(candidate.fsPath, 'utf-8');
+			const json = JSON.parse(text);
+			return json;
+		} catch (error) {
+			console.warn(`Failed to read ${candidate}:`, error);
```
aeschli: remove console statement

src/services/cssNavigation.ts
```diff
@@ -458,27 +475,109 @@ export class CSSNavigation {
...
+		return undefined;
+	}
+
+	private async findNearestSettings(workspaceFolder: string, documentUri: string): Promise<URI | undefined> {
+		// Walks up from documentUri toward workspaceFolder. Stop at first .vscode/settings.json
+		const document = URI.parse(documentUri);
+		const root = URI.parse(workspaceFolder);
+		let current = document;
+
+		while (current.path.startsWith(root.path)) {
```
aeschli, line 566: Should be a proper isEqualOrParent
https://github.com/microsoft/vscode/blob/62fb9237b776e53ecfbdac4050dd55c61912e3d8/src/vs/base/common/extpath.ts#L227

aeschli:
Finding and loading settings should not be part of the vscode-css-languageservice. That should be done outside, by the client of the service.
vscode-css-languageservice should just get the AliasSettings

aeschli (changes requested):
Finding and loading settings should not be part of the vscode-css-languageservice. That should be done outside, by the client of the service.
vscode-css-languageservice should just get the AliasSettings

PR #495, opened 2026-07-14
Support the modern attr() type() syntax

### LINKED ISSUE #451: new `attr()` function with `type(<length>)` falsely reports syntax error
Using the redesigned css `attr()` function displays a syntax error:

```css
#foo {
  left: attr(x type(<length>));
}
```

Results in:
* at first `<`: `) expected css(css-rparentexpected)`
* at closing `}`: `at-rule or selector expected css(css-ruleorselectorexpected)`

Details about the spec:
* https://drafts.csswg.org/css-values-5/#attr-notation
* https://developer.chrome.com/blog/advanced-attr
* https://developer.mozilla.org/en-US/docs/Web/CSS/attr

This is pretty new and experimental feature in CSS, so I understand why it's not supported. Happy to try to implement if help is welcome.

ATKasem:
Fixes #451

`attr()` is parsed through the generic function machinery, so the modern `attr( <name> <type>?, <fallback>? )` form reported false syntax errors. `attr(data-x type(<length>))` produced `) expected` and `at-rule or selector expected` because the `<length>` syntax tokens inside `type()` aren't valid value terms.

This teaches the parser the `type( <syntax> )` argument (`*`, `<type-name>` components joined with `|`, with optional `+`/`#` multipliers), scoped to inside `attr()` via a small depth counter so a user-defined `type()` function elsewhere (e.g. a Sass `@function type()`) is left alone.

Everything that already parsed keeps parsing the same way — legacy `attr(name)` / `attr(name unit)` / `attr(name, fallback)`, `raw-string`, `type(*)`, empty `attr()`, and Sass/Less variables. Malformed types such as `attr(x type())` or `attr(x type(<length))` still report an error.

Added parser tests in `parser.test.ts` covering the legacy forms, the modern `type()` forms, the non-attr `type()` guard, and the malformed cases.

### DIFF
--- a/src/parser/cssParser.ts
+++ b/src/parser/cssParser.ts
@@ -29,6 +29,9 @@ export class Parser {
 
 	private lastErrorToken?: IToken;
 
+	// Tracks nesting inside `attr()`, where `type(<syntax>)` is a valid argument.
+	protected attrDepth: number = 0;
+
 	constructor(scnr: Scanner = new Scanner()) {
 		this.scanner = scnr;
 		this.token = { type: TokenType.EOF, offset: -1, len: 0, text: '' };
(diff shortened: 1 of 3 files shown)


ATKasem:

aeschli (approved):

aeschli:
Thanks @ATKasem !

Giuspepe (approved):

lszomoru (approved):

PR #498, opened 2026-07-15
Fix completion suggestions inside comments

### LINKED ISSUE #429: Pseudo-class/element suggestions are triggered by colon (`:`) character in comment contexts
When I type the colon character (`:`) in a single or multi-line comment in a CSS file, VS Code displays suggestions to autocomplete with CSS pseudo-elements and pseudo-classes, which is not expected in a comment. See screenshot below:

Image

I've opened the issue here suggested by @starball5 in https://stackoverflow.com/a/79525038/10015695.

NgoQuocViet2001:
Fixes #429.

## Summary

- stop completion calculation when the cursor is inside a comment
- use the CSS, SCSS, or LESS scanner so both block and single-line comments are handled
- keep selector completion available immediately after a closed block comment

## Testing

- `npm run compile`
- full test suite: 710 passed

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -18,6 +18,9 @@ import * as l10n from '@vscode/l10n';
 import { isDefined } from '../utils/objects.js';
 import { CSSDataManager } from '../languageFacts/dataManager.js';
 import { PathCompletionParticipant } from './pathCompletion.js';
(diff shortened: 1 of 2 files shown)


aeschli:
duplicate of https://github.com/microsoft/vscode-css-languageservice/pull/506

PR #499, opened 2026-07-15
Add place-items value completions

### LINKED ISSUE #455: CSS `place-items` code completion
It looks like code completions are missing for [`place-items`](https://developer.mozilla.org/en-US/docs/Web/CSS/place-items)

NgoQuocViet2001:
Fixes #455.

## Summary

- resolve value completions for two-property shorthands whose syntax references the longhand properties
- use `align-items` values for the first `place-items` value and `justify-items` values for the optional second value
- avoid copying generated CSS data into a special case

## Testing

- `npm run compile`
- full test suite: 710 passed

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -344,7 +344,7 @@ export class CSSCompletion {
 					}
 				}
 			}
-			this.getValueEnumProposals(entry, existingNode, result);
+			this.getValueEnumProposals(this.getValueEnumEntry(entry, node), existingNode, result);
 			this.getCSSWideKeywordProposals(entry, existingNode, result);
 			this.getUnitProposals(entry, existingNode, result);
 		} else {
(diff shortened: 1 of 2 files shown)


src/services/cssCompletion.ts
```diff
@@ -344,7 +344,7 @@ export class CSSCompletion {
 					}
 				}
 			}
-			this.getValueEnumProposals(entry, existingNode, result);
```
aeschli, line 347: Can the getValueEnumEntry function call go inside getValueEnumProposals ?

aeschli (changes requested):

src/services/cssCompletion.ts
NgoQuocViet2001: Yes — moved it inside in `be92e8c`.

The one constraint was the other caller: `getCompletionsForMediaCondition` calls `getValueEnumProposals(media, null, result)` with an `IDescriptorData` and no declaration, so the declaration had to be optional rather than required:

```ts
public getValueEnumProposals(entry: IPropertyData | IDescriptorData, existingNode: nodes.Node | null, result: CompletionList, declaration?: nodes.Declaration): CompletionList {
    if (declaration) {
        entry = this.getValueEnumEntry(entry, declaration);
    }
    if (entry.values) {
```

`getValueEnumEntry` widened to the same `IPropertyData | IDescriptorData` union the method already took — `IDescriptorData` carries `syntax` and `values` too, so nothing in its body needed changing. The public signature is unchanged for existing callers.

`npm run compile` (tsc + eslint) is clean and the full suite is 710 passed. Reverting just the resolution block fails the new `place-items` case with `center should only existing once`, so it is still pinned by the test rather than by where the call sits.

Happy to make the parameter required and pass `null` from the media path instead if you would rather not have an optional there.

PR #506, opened 2026-08-20
No completions inside comments

### LINKED ISSUE #429: Pseudo-class/element suggestions are triggered by colon (`:`) character in comment contexts

kakiuwang-ui:
Fixes #429

### Problem

Comments are trivia and never make it into the node tree, so `doComplete` resolves the node path as if the comment were not there. Typing inside a comment therefore offers the completions of the *surrounding* context:

```css
/* foo: */            ← at-rules and selectors
.foo { /* color: */ } ← the full property list
```

The symptom reported in #429 is the most visible one: typing a `:` anywhere in a comment pops up the entire pseudo-class/pseudo-element list. It applies to SCSS/LESS `// ...` line comments as well.

### Fix

Locate comments by scanning the document (`Scanner` with `ignoreComment = false`) and return an empty list when the position is inside one. `SCSSCompletion`/`LESSCompletion` override `getScanner` so `//` line comments are recognized in those dialects.

Two details worth calling out:

- The scan mirrors the parser's `inURL` handling. Without it, `//` in an unquoted `url(http://server/a.png)` scans as a line comment in SCSS/LESS and would have suppressed completions for the rest of the line. There are tests for both the unquoted and the protocol-relative (`url(//server/a.png)`) form.
- A position at the very end of a comment that was never closed (`/* ...` at EOF, or `// ...` at EOL) still counts as inside it, while a position right after a closing `*/` does not.

The scan stops as soon as it passes the requested offset, so it does not walk the whole document.

### Tests

Added `no completions in comments` / `completions next to comments` to the CSS, SCSS and LESS completion tests.

One test-infra change was needed: `testCompletionFor` built its position with `Position.create(0, offset)`, which silently clamps to the first line and made multi-line test inputs untestable. It now uses `document.positionAt(offset)` — identical for single-line inputs, which is every existing case.

All 722 tests pass.

### DIFF
--- a/src/services/cssCompletion.ts
+++ b/src/services/cssCompletion.ts
@@ -68,21 +68,50 @@ export class CSSCompletion {
 	}
 
 	/**
+	 * A node never starts inside a comment, so the last node starting before `offset` is a safe
+	 * place to start scanning from: a comment holding `offset` cannot have started any earlier.
+	 */
+	private getCommentScanStart(styleSheet: nodes.Stylesheet, offset: number): number {
+		let node: nodes.Node | null = styleSheet;
+		let start = 0;
+		while (node) {
+			let last: nodes.Node | null = null;
+			for (const child of node.getChildren()) {
+				if (child.offset >= 0 && child.offset < offset) {
+					last = child; // children are in document order, so the last match wins
+				}
+			}
+			if (last) {
+				start = last.offset;
+				// `url(...)` is scanned in a dedicated mode, so never start inside one.
+				const uriLiteral = last.findParent(nodes.NodeType.URILiteral);
+				if (uriLiteral) {
+					return uriLiteral.offset;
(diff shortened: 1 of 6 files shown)


aeschli (changes requested):
Instead of scanning from the beginning of the document we can use the available stylesheet to find the start of a previous declaration and go from there

aeschli (approved):

Giuspepe (approved):

lszomoru (approved):

